PR #1, opened 2014-05-07
Jruby pg

jvshahid:
I thought I should open a pr to get the ball rolling. This still needs some work, there are currently 9 failures that I need to fix. Feedback is welcome, specially with the changes to the Rakefile, etc.


larskanis:
It's a bunch of work incorporated into this PR, but I think the approach to add a JRuby extension to ruby-pg is no longer followed and it can be closed.

@jvshahid I read about better PostgreSQL support for rails-7 on JRuby, but didn't find a link to any code. What is it about? Should I add a pointer to the README of ruby-pg?

PR #4, opened 2014-09-05
Fix segfault when clearing result yielded to notice_receiver

jeremyevans:
This fixes issue #185 in bitbucket (which I reported). In order
to make a double free impossible, you need to make sure that
clearing the notice receiver result doesn't call PQclear. In
order to make a segfault impossible if accessing the notice
receiver result outside of the notice receiver proc, you need
to use rb_ensure to ensure the data pointer is cleared before
returning from the proc call.

This is the simplest implementation I could think of, using a
result subclass for the notice receiver results, where clear
doesn't call PQclear.

### DIFF
--- a/ext/pg.h
+++ b/ext/pg.h
@@ -115,6 +115,7 @@ extern VALUE rb_eConnectionBad;
 extern VALUE rb_mPGconstants;
 extern VALUE rb_cPGconn;
 extern VALUE rb_cPGresult;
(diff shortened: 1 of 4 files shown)


jeremyevans:
Since #6 solved this in a different way, I think this can be closed.  If you still think it is an issue, please reopen.

larskanis:
It can be closed - thank you!

PR #5, opened 2014-09-28
Replaced StringValuePtr to StringValueCStr

ymmtmdk:
On ruby-head environment, I got the following error: 

`PG::CharacterNotInRepertoire: ERROR:  invalid byte sequence for encoding "UTF8": 0xe0 0x0c 0xf7`

I researched the code and found it may occured by not terminated C string.
In Ruby’s README.ext,
```
You can also use the macro named StringValueCStr(). This is just
like StringValuePtr(), but always add nul character at the end of
the result. If the result contains nul character, this macro causes
the ArgumentError exception.
StringValuePtr() doesn't guarantee the existence of a nul at the end
of the result, and the result may contain nul.
```

So, I made this patch and solved the error.
Is this right?

### DIFF
--- a/ext/pg.h
+++ b/ext/pg.h
@@ -55,9 +55,9 @@
 #	define RSTRING_PTR(x) RSTRING((x))->ptr
 #endif /* RSTRING_PTR */
 
-#ifndef StringValuePtr
-#	define StringValuePtr(x) STR2CSTR(x)
-#endif /* StringValuePtr */
+#ifndef StringValueCStr
+#	define StringValueCStr(x) STR2CSTR(x)
+#endif /* StringValueCStr */
 
 #ifdef RUBY_18_COMPAT
 #	define rb_io_stdio_file GetWriteFile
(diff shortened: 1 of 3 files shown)


meoooh:
Is there any update?

larskanis:
This is fixed in commit https://bitbucket.org/ged/ruby-pg/commits/f0b7f99 and released as part of pg-0.18.0.

meoooh:
thank you! @larskanis

mtarnovan:
I'm still seeing this errors in a Rails up using 0.18.1. Any ideas?

JacobEvelyn:
Same here.

larskanis:
@mtarnovan , @JacobEvelyn could you please put a stacktrace of the failure into a gist?

JacobEvelyn:
Looks like I was only getting this error when trying to save binary data to a `VARCHAR` field rather than a `BYTEA`.

PR #7, opened 2015-02-06
Use fixed size arrays in case the compiler doesn't support C99 variable ...

larskanis:
...length arrays - notably the MSVC compiler.

This is an alternative to https://bitbucket.org/ged/ruby-pg/pull-request/20 .

Note: The MSVC build environment is currently not tested on a regular base,
so is not fully supported.

### DIFF
--- a/ext/extconf.rb
+++ b/ext/extconf.rb
@@ -94,6 +94,10 @@ $defs.push( "-DHAVE_ST_NOTIFY_EXTRA" ) if
 have_header 'unistd.h'
 have_header 'ruby/st.h' or have_header 'st.h' or abort "pg currently requires the ruby/st.h header"
 
+checking_for "C99 variable length arrays" do
+	$defs.push( "-DHAVE_VARIABLE_LENGTH_ARRAYS" ) if try_compile('void test_vla(int l){ int vla[l]; }')
+end
+
 create_header()
 create_makefile( "pg_ext" )
 
--- a/ext/pg.h
+++ b/ext/pg.h
@@ -133,6 +133,15 @@
 typedef long suseconds_t;
 #endif
 
+#if defined(HAVE_VARIABLE_LENGTH_ARRAYS)
+	#define PG_VARIABLE_LENGTH_ARRAY(type, name, len, maxlen) type name[(len)];
+#else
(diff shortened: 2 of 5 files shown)


lowjoel:
How about testing for alloca instead, and only using the constant size array when both variable length arrays and alloca are not available?

GCC supports alloca, so while it isn't in POSIX, anywhere else where the extension is compiled using GCC alloca should be available.

larskanis:
Variable length arrays are deallocated at the end of the block but alloca() at the end of the function (as you already pointed out). This makes it unusable as a simple preprocessor macro. We would need to use different code paths (#ifdefs or several macros), to implement both ways at all places which makes use of variable length arrays. The same is true for alternatives based on malloc/free or malloca/freea.

lowjoel:
yes, hence my PR on Bitbucket tended to move all variable declarations to the start of the function. The functionality is equivalent on both compilers supporting alloca and those which support VLAs.

I'm against using hardcoded lengths where possible. Indeed while MSVC isn't tested regularly, having this compiling fine might allow people to use in production, giving a potential buffer overflow vector, and a false sense of security.

There are multiple considerations here. I think it is more important to have correct support than having seemingly-correct support.

larskanis:
@lowjoel the array size is checked - did you read the sources [here](https://github.com/ged/ruby-pg/pull/7/files#diff-3de60090228d88f5e41cc4cf8ac84dd6R140) ? Does it work with MSVC?

lowjoel:
Right, I glossed over that. If that's the case, then I have no complaints.

I have not built it yet. I'm not sure if it would work since there's a conditional in the variable initialiser.

lowjoel:
@larskanis thanks! Your branch compiles.

If I were to set up AppVeyor, how would the account be handled? Their config is a bit different from how Travis does it.

larskanis:
Never used Appveyor myself, but I think something like forking ruby-pg (github or bitbucket), connecting to Appveyor, do necessary additions/changes to that fork and sending a pull request should work. The test itself should include a build and a test suite run. The PostgreSQL server and Ruby environment could be any supported version, since we do regular testing against oldest/newest supported versions on travis-ci already. @ged or I will add the Appveyor-connection to our accounts, to ensure it is tested on the MSVC platform.

PR #9, opened 2015-06-22
Date decoder to return real Dates

ramfjord:
I'm not sure if you guys are into this idea, but if you are here's a pull request.  I would love to have this personally as it removes some somewhat hacky workarounds I have to do.

### DIFF
--- a/lib/pg/text_decoder.rb
+++ b/lib/pg/text_decoder.rb
@@ -7,7 +7,7 @@ module PG
 
 			def decode(string, tuple=nil, field=nil)
 				if string =~ ISO_DATE
-					Time.new $1.to_i, $2.to_i, $3.to_i
+					::Date.new $1.to_i, $2.to_i, $3.to_i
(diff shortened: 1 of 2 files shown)


cbandy:
N.B. there is no `Date` in Ruby core.

ramfjord:
I see.  I'm unfamiliar with projects that require such widespread support.  Is using ``stdlib`` classes a non-starter then?  Including any solution to [adding json support](https://bitbucket.org/ged/ruby-pg/issue/212/json-jsonb-default-type-map-for-results) using JSON?

larskanis:
@ramfjord Sorry I missed your request. Rails also casts PostgreSQLs dates to Date objects, and it seems to be reasonable for me. We break backward compatibility, but I guess it is not (yet) in wide use. Stdlib is OK, but the require is missing - this is the reason why travis-ci failed. What is the reason for closing the PR?

PR #10, opened 2015-09-08
Fixing compilation problems with Microsoft Visual Studio 2008.

jfali:
- inttypes.h is not available in MSVC 2008
- Structs members must be initialized implicitly


larskanis:
Merged! But, broken compilers bother...

PR #11, opened 2016-01-09
Respect character encoding of all strings sent to the server.

larskanis:
Previously all strings sent to the server were sent in their internal binary representation, without respecting the character encoding of strings. Now the encoding of all strings is compared with the current connection encoding and converted if they are different.

Since coders are independent from any database connection, this adds a second parameter to `PG::Coder#encode`, that allows to define the destination encoding, which previously was always ASCII_8BIT. This encoding should be set to the connection encoding, in practice.

This also adds a lot of tests regarding the character encoding while encoding and decoding data.

This implements issue #231 : https://bitbucket.org/ged/ruby-pg/issues/231

@ged Any objections to merge this for 0.19.0 ?

### DIFF
--- a/ext/pg.h
+++ b/ext/pg.h
@@ -205,7 +205,7 @@ typedef struct {
 } t_pg_result;
 
 
-typedef int (* t_pg_coder_enc_func)(t_pg_coder *, VALUE, char *, VALUE *);
+typedef int (* t_pg_coder_enc_func)(t_pg_coder *, VALUE, char *, VALUE *, int);
 typedef VALUE (* t_pg_coder_dec_func)(t_pg_coder *, char *, int, int, int, int);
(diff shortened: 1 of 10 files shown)


ged:
I somehow missed this over the holidays. Looks good to me! :+1:

ghidinelli:
Thanks for doing this work @larskanis - we hit this problem occasionally and would love to have a fix. 👍 

Prior to 0.19 being released, is there something in Ruby we can do prior to sending the data to Postgres to emulate this encoding change?

larskanis:
Does the pg pre-release work for you? With pg < 0.19 you need to call String#encode on all strings before transmission to the server.

larskanis:
This is merged to master: https://github.com/ged/ruby-pg/commit/e5cb1df93c496560444941d7a531e4cd675ec7c2

PR #14, opened 2016-07-22
Include ext/errorcodes.def so this gem can be installed with bundler.

gfortaine:


larskanis:
Thank you for your pull request! This is on the master branch, now: https://github.com/ged/ruby-pg/commit/367d4df410c505020e0322367b236bb8a77eaf4c

PR #17, opened 2017-09-05
Check if the argument for #Connection is an URI

### LINKED ISSUE #265: Colon not allowed in a connection string password
**[Original report](https://bitbucket.org/ged/ruby-pg/issue/265) by Jan Deelstra (Bitbucket: [jjoos](https://bitbucket.org/jjoos), GitHub: [jjoos](https://github.com/jjoos)).**

----------------------------------------

According to the documentation only single quotes and backslashes need to be escaped ( https://www.postgresql.org/docs/current/static/libpq-connect.html#LIBPQ-CONNSTRING ). This seemed to work correctly in older versions (up to 0.18, but in newer versions it doesn't work anymore. It doesn't matter if the password is contained within single quotes or not. Normally I would have worked around this issue by using a different method signature, but I'm reusing these connection strings for `db_link` connections which makes it pretty inconvenient.

PG 0.18: (correct behaviour)
```
[1] pry(main)> require 'pg'
=> false
[2] pry(main)> PG::Connection.connect("host=localhost port=5432 dbname=looker user=postgres password=a:a")
PG::ConnectionBad: could not connect to server: Connection refused
	Is the server running on host "localhost" (::1) and accepting
	TCP/IP connections on port 5432?
could not connect to server: Connection refused
	Is the server running on host "localhost" (127.0.0.1) and accepting
	TCP/IP connections on port 5432?
from (pry):2:in `initialize'
```
PG 0.19 and later:
```
[1] pry(main)> require 'pg'
=> false
[2] pry(main)> PG::Connection.connect("host=localhost port=5432 dbname=looker user=postgres password=a:a")
URI::InvalidURIError: bad URI(is not URI?): host=localhost port=5432 dbname=looker user=postgres password=a:a
from /Users/jjoos/.rbenv/versions/2.3.3/lib/ruby/2.3.0/uri/rfc3986_parser.rb:67:in `split'
```

This bug is in `/lib/pg/connection.rb`:
```
class PG::Connection
       ...
	def self::parse_connect_args( *args )
                ...
		if args.length == 1
			case args.first
			when URI, URI.regexp
				uri = URI(args.first)
				options.merge!( Hash[URI.decode_www_form( uri.query )] ) if uri.query
			when /=/
				# Option string style
				option_string = args.first.to_s
                        ...
```
Since:
```
!!(URI.regexp.match "host=localhost port=5432 dbname=looker user=postgres password='a:a'")
=> true
```
the code thinks it's dealing with an uri connection string, because it checks if the string contains an uri and not if it **is** an uri. The regex is missing a start and end anchor, changing the code like this will fix the issue:
```
			when URI, /\A#{URI::regexp}\z/
```
since:
```
!!(/\A#{URI::regexp}\z/.match "host=localhost port=5432 dbname=looker user=postgres password='a:a'")
=> false
```

jjoos:
Previously the code checked if the argument
contained an URI instead of being one. This
caused some connection strings being mis-
dected as connection URI's.

fixes #265  ( https://bitbucket.org/ged/ruby-pg/issues/265/colon-not-allowed-in-a-connection-string )

### DIFF
--- a/lib/pg/connection.rb
+++ b/lib/pg/connection.rb
@@ -47,7 +47,7 @@ class PG::Connection
 
 		if args.length == 1
 			case args.first
-			when URI, URI.regexp
+			when URI, /\A#{URI.regexp}\z/
 				uri = URI(args.first)
 				options.merge!( Hash[URI.decode_www_form( uri.query )] ) if uri.query
 			when /=/
--- a/spec/pg/connection_spec.rb
+++ b/spec/pg/connection_spec.rb
@@ -119,6 +119,19 @@ describe PG::Connection do
 	end
 
 	it "connects successfully with connection string" do
+		conninfo_with_colon_in_password = "host=localhost user=a port=555 dbname=test password=a:a"
+
+		string = described_class.parse_connect_args( conninfo_with_colon_in_password )
+
+		expect( string ).to be_a( String )
+		expect( string ).to match( %r{(^|\s)user=a} )
+		expect( string ).to match( %r{(^|\s)password=a:a} )
+		expect( string ).to match( %r{(^|\s)host=localhost} )
+		expect( string ).to match( %r{(^|\s)port=555} )
+		expect( string ).to match( %r{(^|\s)dbname=test} )
+	end
+
+	it "connects successfully with connection string" do
 		tmpconn = described_class.connect( @conninfo )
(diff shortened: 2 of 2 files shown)


larskanis:
LGTM - merged. Thank you!

PR #19, opened 2018-01-02
Added a C implementation of the SimpleDecoder's timestamp decode functions

remy-cassia-tech:
- SimpleDecoder::TimestampWithTimeZone::decode
- SimpleDecoder::TimestampWithoutTimeZone::decode

### DIFF
--- a/ext/pg_text_decoder.c
+++ b/ext/pg_text_decoder.c
@@ -28,10 +28,26 @@
  *
  */
 
+#include "ruby/version.h"
 #include "pg.h"
 #include "util.h"
 #ifdef HAVE_INTTYPES_H
 #include <inttypes.h>
+#endif
+
+#ifdef RUBY_API_VERSION_MAJOR
+#       if RUBY_API_VERSION_MAJOR > 2 || (RUBY_API_VERSION_MAJOR == 2 && RUBY_API_VERSION_MINOR >= 3)
+                /* use C implementation of the SimpleDecoder`s timestamp function
+                 * when using ruby >= 2.3 */
+#               define PG_TEXT_DECODER_TIMESTAMP_EXT
+#       endif
+#endif
+
+#ifdef PG_TEXT_DECODER_TIMESTAMP_EXT
+#       include <ctype.h>
+#       include <time.h>
+static VALUE rb_cPG_TimestampWithTimeZone;
+static VALUE rb_cPG_TimestampWithoutTimeZone;
 #endif
 
 VALUE rb_mPG_TextDecoder;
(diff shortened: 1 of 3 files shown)


ext/pg_text_decoder.c
```diff
@@ -418,4 +608,14 @@ init_pg_text_decoder()
 	pg_define_coder( "Array", pg_text_dec_array, rb_cPG_CompositeDecoder, rb_mPG_TextDecoder );
 	/* dummy = rb_define_class_under( rb_mPG_TextDecoder, "FromBase64", rb_cPG_CompositeDecoder ); */
 	pg_define_coder( "FromBase64", pg_text_dec_from_base64, rb_cPG_CompositeDecoder, rb_mPG_TextDecoder );
+
+#ifdef PG_TEXT_DECODER_TIMESTAMP_EXT
+        /* Document-class: PG::TimestampWithTimeZone < PG::SimpleDecoder */
+        rb_cPG_TimestampWithTimeZone = rb_define_class_under( rb_mPG_TextDecoder, "TimestampWithTimeZone", rb_cPG_SimpleDecoder );
+        rb_define_method( rb_cPG_TimestampWithTimeZone, "decode", pg_text_dec_timestamp_with_time_zone, -1 );
+
```
larskanis: Why didn't you use `pg_define_coder()` here?

ext/pg_text_decoder.c
```diff
@@ -392,6 +408,180 @@ pg_text_dec_from_base64(t_pg_coder *conv, char *val, int len, int tuple, int fie
...
+        int gmt_offset;
+
+        gmt_offset = tz_hour * 3600 + tz_min * 60 + tz_sec;
+        if (tz_neg)
+        {
+          gmt_offset = - gmt_offset;
+        }
+        ts.tv_sec = time - gmt_offset;
+        ts.tv_nsec = nsec;
+        return rb_time_timespec_new(&ts, gmt_offset);
```
larskanis: Couldn't we use `rb_funcall()` to Time.new as fallback for Ruby<2.3? This way the Regexp version could be removed entirely.

ext/pg_text_decoder.c
```diff
@@ -392,6 +408,180 @@ pg_text_dec_from_base64(t_pg_coder *conv, char *val, int len, int tuple, int fie
...
+
+static int str2_to_int(const char *str)
+{
+  return char_to_digit(str[0]) * 10
+       + char_to_digit(str[1]);
+}
+
+static VALUE pg_text_decoder_timestamp_do(VALUE rstr, int with_timezone)
+{
+  const char *str = StringValuePtr(rstr);
```
larskanis: `StringValuePtr()` doesn't ensure, that the corresponding string is zero-terminated. Therefore `RSTRING_LEN()` should be checked as well.

ext/pg_text_decoder.c
remy-cassia-tech: I added a commit to use pg_define_coder: a65a71e40909e6ddb6c4de57369390935066ea79
But I have a question. The decode function does not have the original Ruby string as an argument. In case of a malformed string, I can't return the original string as before. I changed the code to raise an exception as other decode functions do. What do you think ?

ext/pg_text_decoder.c
remy-cassia-tech: You`re right. Using the pg_define_coder function resolves this problem since the caller ensures that the data is zero terminated

remy-cassia-tech: I just checked the `pg_coder_decode` function and it does *not* ensure the string is zero-terminated as the comment in the `pg_decoder.c` file states.
Either the `pg_coder_decode` function should ensure the string is zero-terminated or the comment in the `pg_decoder.c` file should be updated. The `pg_text_dec_float` function assumes the string is zero-terminated too.
What do you think is better, update the pg_coder_decode function or update the comment?

ext/pg_text_decoder.c
```diff
@@ -392,6 +396,204 @@ pg_text_dec_from_base64(t_pg_coder *conv, char *val, int len, int tuple, int fie
...
+          INT2NUM(mon),
+          INT2NUM(day),
+          INT2NUM(hour),
+          INT2NUM(min),
+          sec_value,
+          gmt_offset_value);
+#endif
+    }
+  }
+  return rb_tainted_str_new(rstr, len);
```
larskanis: The fallback should be `pg_text_dec_string()`.

remy-cassia-tech:
First, thanks for reviewing this branch. Using this C implementation improved a lot our performances.
I didn't wrote new test cases. I only relied on the existing ones.
Tomorrow I'll add a bunch of new test cases

ext/pg_text_decoder.c
```diff
@@ -392,6 +396,204 @@ pg_text_dec_from_base64(t_pg_coder *conv, char *val, int len, int tuple, int fie
...
+        tz_sec = str2_to_int(str);
+        str += 2;
+      }
+    }
+    if (*str == '\0')
+    {
+#if RUBY_API_VERSION_MAJOR > 2 || (RUBY_API_VERSION_MAJOR == 2 && RUBY_API_VERSION_MINOR >= 3)
+      // must have consumed all the string
+      struct tm tm;
+      tm.tm_year = year - 1900;
```
larskanis: What about timestamps before 1900?

remy-cassia-tech: In C, the tm_tm_year field stores the number of years since 1900, but this number can be negative.
I added a test that check that date before 1823 and after 2116 are supported

larskanis:
I did some minor changes on [this branch](https://github.com/larskanis/ruby-pg/commits/vhl-time-decoder). Unfortunately some specs [fail on Windows](https://ci.appveyor.com/project/larskanis/ruby-pg-xa3f5/build/1.0.94) while all specs [succeed on linux](https://travis-ci.org/larskanis/ruby-pg). Any ideas?

remy-cassia-tech:
From what I can see in the log, specs are failing when using x64-mingw32.
Specs succeed when using Travis CI and i386-mingw32.
The specs are failing because the decode function returned the original string.
The only way to fail is if the `mktime` function returns an error (I don't think it can be an error while parsing the string)
I read the mktime documentation again, and I think the tm.tm_isdst should be set to -1 instead of 0 (a positive value means DST is in effect; zero means that DST is not in effect; and a negative value means that mktime() should (use timezone information and system databases to) attempt to determine whether DST is in effect at the specified time).
I looked at the mingw implementation of `mktime` and wrongly setting tm.tm_isdst can result in an error.
I don't have access to a x64-mingw32 platform. Could you do this test ?

larskanis:
Thanks for your investigation! Unfortunately I couldn't find a way to convert all dates on Windows error-free. `tm.tm_isdst` didn't make a difference. I therefore switched Windows to the path of Time.new : https://github.com/larskanis/ruby-pg/commits/vhl-time-decoder

alboyadjian:
@larskanis Do you think there's any chance of this getting merged?  The performance boost is pretty noticeable, and we'd like to avoid using a forked repo.

larskanis:
I did some more steps towards merging this PR.

1. I [extracted the specs](https://github.com/ged/ruby-pg/commit/d40cff15ff8c403439110b24dc7e18f9f28ce01f) and merged it to master, so that the expected behavior is well defined, now.

2. I did some more testing and noticed, that timegm() fails on Android (and probably most of non glibc systems as well). Such a portability issue is a no-go for me. I therefore [reverted the fast path](https://github.com/larskanis/ruby-pg/commit/b66cec768a368d95865222e608ca73fc32d13270) in [my branch](https://github.com/larskanis/ruby-pg/commits/vhl-time-decoder).

3. I did some very simple benchmarks with this command and got the following results:
```sh
ruby -Ilib -rpg -e "deco = PG::TextDecoder::TimestampWithoutTimeZone.new; tstr='2016-01-02 23:23:59.0'; st=Time.now; 100000.times{ deco.decode(tstr) }; p Time.now-st"
```

version|time
-------|-----
master (RegExp)|1.18s
vhl-time-decoder slow path|0.74s
vhl-time-decoder fast path|0.19s

The times when running the conversion as part of conn.exec are probably slightly different, but I wonder whether the performance improvement is enough to merge the vhl-time-decoder slow path?

@ged Do you have an opinion here?

ged:
I haven't had a chance to look at it in depth yet, but I will this weekend. Just didn't want you to think I was ignoring ya. :)

larskanis:
I added an updated version of this PR as #20.

larskanis:
This is merged as part of #20 . Thank you for this contribution!

PR #20, opened 2018-04-30
Faster time decoder

larskanis:
This moves PG::SimpleDecoder::TimestampWithTimeZone and PG::SimpleDecoder::TimestampWithoutTimeZone to C. It is based on https://github.com/ged/ruby-pg/pull/19 with some tweaking by me.

I tested the implementation successfully on Windows-10 with MINGW (RubyInstaller), Ubuntu-18.04 with glibc, Alpine-Linux-3.7 with musl-libc and Android-7.0 with bionic-libc (termux).

The conversion is done by a fast path and as a fallback by a slower path. The slow path is used under these conditions:
* `timegm()` is not available (on Windows)
* `RUBY_VERSION < 2.3.0` (it doesn't provide `rb_time_timespec_new()`)
* time conversion by `timegm()` or `mktime()` fails at runtime (happens on Android)

I did some very simple benchmarks with the following results:

```
ruby -Ilib -rpg -e "map=PG::TypeMapByColumn.new([PG::TextDecoder::TimestampWithTimeZone.new]); res=PG.connect.exec(\"SELECT '2016-01-02 22:23:59.3-04'::timestamptz\").map_types!(map); st=Time.now; 10000.times{ res.getvalue(0,0) }; p Time.now-st"
```

version|time
-------|-----
master (RegExp)|0.71s
vhl-time-decoder slow path|0.41s
vhl-time-decoder fast path|0.022s

### DIFF
--- a/ext/extconf.rb
+++ b/ext/extconf.rb
@@ -76,6 +76,7 @@ have_func 'PQsetSingleRowMode' or
 have_func 'PQconninfo'
 have_func 'PQsslAttribute'
 have_func 'PQencryptPasswordConn'
+have_func 'timegm'
 
 have_const 'PG_DIAG_TABLE_NAME', 'libpq-fe.h'
 
--- a/ext/pg_text_decoder.c
+++ b/ext/pg_text_decoder.c
@@ -535,7 +535,7 @@ static VALUE pg_text_decoder_timestamp_do(t_pg_coder *conv, char *val, int len, 
     int tz_hour = 0;
     int tz_min = 0;
     int tz_sec = 0;
- 
+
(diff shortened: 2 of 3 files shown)


larskanis:
@ged Thank you very much for reviewing! I'll merge it when time permits.

Moreover I think it makes sense to add another de+encoder for timestamps without timezone, which are treated as UTC time instead of local time. This is how ActiveRecord handles timestamps without timezone per default.

alboyadjian:
Thank you for merging this! It's a big speed boost for us as we're using the raw exec for a hefty query. Thank you for your cleanup @larskanis

PR #21, opened 2018-05-12
Async API extension and fixes

larskanis:
This overhauls the async API in such a way that more async_* convenience methods are provided. I had this plan since years, but now it was proposed in https://github.com/rails/rails/pull/32820 , so that I took the opportunity to make it happen.

New is also PG::Connection#discard_results, which mimics the discarding of results at the beginning of PQexec(). It can be used together with send_* methods.

It also tests and fixes async_exec for some edge cases with COPY operations.

See the single commit messages for more information to the changes.


larskanis:
Thank you @cbandy ! I fixed the two issues.

spec/pg/connection_spec.rb
```diff
@@ -893,15 +890,36 @@
...
 
-	it "calls the block if one is provided to #async_exec" do
+	it "discards previous errors before waiting on #exec and #async_exec", :without_transaction do
+		@conn.send_query( "ERROR" )
+		res = @conn.exec( "select 43 as one" )
+		expect( res.to_a ).to eq( [{ 'one' => '43' }] )
+	end
+
+	it "discards previous cancels before waiting on #exec and #async_exec", :without_transaction do
+		@conn.send_query( "SEELCT pg_sleep(5)" )
```
matthewd: Is this typo intentional?

matthewd:
The testing strategy in `connection_async_spec.rb` raises another question for me: if the methods work exactly the same for the caller, and the `async_` version is just "better", maybe the naming should be the other way around?

It's maybe possible to imagine some caller who cares that they're calling exactly `PQexecPrepared`, so I can understand retaining a method that definitely does that, but ISTM it needn't necessarily be named `exec_prepared` (vs `sync_exec_prepared`, say). For 99.9% of callers, they want `exec_prepared` to do the thing that `PQexecPrepared` does, but don't care _how_ it does it... if there's an approach that has better behavioural trade-offs in the general case, I think it's right for the gem to make that decision for us.

In short: I think this PR sets us up for a consistent stream of "unless you really have a specific reason, you should be using `async_prepare` instead of `prepare`" issues on downstreams. Instead, I suggest we do the right thing for those callers, and force the one with obscure requirements to adapt to a more specific method name.

spec/pg/connection_spec.rb
larskanis: :smile: Removed the test case - it was a duplication anyway.

larskanis:
@ged is it OK to merge? I would address [#276](https://bitbucket.org/ged/ruby-pg/issues/276/inconsistent-api-for-exec-exec_params-vs) afterwards.

What about the API documentation? I would keep the main description at #exec with links to #sync_exec and #async_exec .

ext/pg_connection.c
```diff
@@ -3125,18 +3155,119 @@ pgconn_get_last_result(VALUE self)
  * This function has the same behavior as #exec,
  * but is implemented using the asynchronous command
  * processing API of libpq.
+ *
+ * Both #exec and #async_exec release the GVL while waiting for server response, so that concurrent threads will get executed.
+ * However #async_exec has two advantages:
+ *
+ * 1. #async_exec can be aborted by signals (like Strg-C), while #exec blocks signal processing until the query is answered.
```
ged: Is `Strg-C` equivalent to `Ctrl-C`?

lib/pg/connection.rb
```diff
@@ -268,5 +268,24 @@ def ssl_attributes
...
+
+	def self.async_api=(enable)
+		REDIRECT_METHODS.each do |ali, (async, sync)|
+			remove_method(ali) if method_defined?(ali)
+			alias_method( ali, enable ? async : sync )
+		end
+	end
+
+	# pg-1.1.0+ defaults to libpq's async API for query related blocking methods
+	self.async_api = true
```
ged: Wouldn't this require bumping the major version? While this isn't a change in the method signature, it definitely changes the behavior of methods with the same name. I think this is a good change going forward, but I think it's significant enough to warrant a major-version change.

matthewd: > it definitely changes the behavior of methods

Does it? My argument was precisely that it doesn't. It changes the _implementation_, but not any ruby- (or postgres-) observable behaviour.

larskanis: I understand it as a bug, if the methods change their behavior due to this PR. That's why I enabled the test suite to run all specs through both sync and async versions of the methods.

There are two (known) differences between sync and async:

1. Signals are no longer delayed until the method returns.
2. The trace outputs are slightly different.

I would argue that delaying signals is a bug, not a feature, so that I think 1. is a fix rather than a behavioral change.

The trace output is an implementation specific feature. We already had a change between PostgreSQL 8.4 and 9.0, so that I count that as internal only. In fact this is the only behavioral change I did to the specs to make it pass.

We aren't in hurry with this change. So I still would like to test pg in various contexts before releasing, to make sure or test suite isn't fooling us. But in the end I would release it as pg-1.1.0.

ged: I guess it depends on what you consider the "API" with regards to semantic versioning. Don't misunderstand me: this change is unquestionably making the library better, and I would advocate for merging it even if it meant a major version change. I agree that the trace output is only revealing internal details which shouldn't have been counted on. I agree that delaying signals is not a feature, but it still is behavior which may change how some systems run. It's definitely on the fringes of what I'd consider a library's "API", but I thought the question still worth asking. If you both are unconcerned, however, I'm fine with just doing careful testing and releasing it as 1.1.0.

larskanis:
[Bitbucket issue #6](https://bitbucket.org/ged/ruby-pg/issues/6/proposed-new-async-api) is related to this PR. Obviously some of the ideas we implement now are already 10 years old.

PR #24, opened 2018-06-17
Add a text decoder for inet/cidr written in C

jeremyevans:
This also includes a pure-ruby text encoder for IPAddr.
It registers the inet/cidr decoder in the basic type map.

Parsing of inet/cidr values is performed using the standard
inet_pton C function.  If there is a mask (optional for inet,
always present for cidr), it is removed before passing the
char* to inet_pton.

As the IPAddr internals haven't changed in basically forever
(at least since 1.8), and for best performance we want to avoid
duplicative checks that would be much slower in ruby, and the
fact that the ipaddr doesn't offer a fast way to create IPAddr
instances with masks, this uses rb_obj_alloc+rb_ivar_set to
create the IPAddr instances if an IPAddr.new instance uses
the @addr, @mask_addr, and @family instance variables.  Any
other instance variables will trigger the more standard ruby
way of calling rb_class_new_instance and then #mask on the
IPAddr instance, which is more than 2x slower.  I also tried
using #mask! (a protected method) instead of #mask, but it
was only a little faster.

This tries to detect big-endian vs. little endian and handle
things correctly.  I'm fairly sure it is correct for IPv4, not
as sure for IPv6, and I don't have a big endian machine to test
on.

This precomputes all possible mask values and stores them in
a couple ruby arrays, so that mask values don't need to be computed
at runtime.  This should allocate the following ruby objects
per decoding:

amd64/ipv4: 1 IPAddr
amd64/ipv6: 1 IPAddr/0-4 Bignum (depending on IP value)
i386/ipv4: 1 IPAddr/0-1 Bignum (depending on IP value)
i386/ipv6: 1 IPAddr/0-4 Bignum (depending on IP value)

Specs for IPAddr encoding and inet/cidr decoding are included.

This was tested and benchmarked with the following code, showing
over a 10x increase in performance compared a the pure ruby
example code shown at the top of lib/pg/basic_type_mapping.rb:

```ruby
require 'pg'
require 'ipaddr'
require 'benchmark/ips'

inet_max = 2**32
inet6_max = 2**128

log_cols = ((ENV['BENCH_LOG_COLS'] || 6).to_i)
log_rows = ((ENV['BENCH_LOG_ROWS'] || 6).to_i)
cols = 2**log_cols

raise "BENCH_LOG_COLS must be <= 10" if log_cols > 10
raise "BENCH_LOG_ROWS must be <= 10" if log_rows > 10

if ENV['PURE_RUBY'] == '1'
  class InetDecoder < PG::SimpleDecoder
    def decode(string, tuple=nil, field=nil)
      IPAddr.new(string)
    end
  end
  class InetEncoder < PG::SimpleEncoder
    def encode(ip_addr)
      ip_addr.to_s
    end
  end
  PG::BasicTypeRegistry.register_type(0, 'inet', InetEncoder, InetDecoder)
end

conn = PG.connect
unless ENV['ALL_STRINGS'] == '1'
  conn.type_map_for_queries = PG::BasicTypeMapForQueries.new conn
  conn.type_map_for_results = PG::BasicTypeMapForResults.new conn
end

conn.exec("BEGIN")
at_exit{conn.exec("ROLLBACK")}
conn.exec("CREATE TABLE inet4_test (#{(0...cols).map{|i| "i#{i} inet DEFAULT '#{IPAddr.new((rand*inet_max).to_i, Socket::AF_INET)}/#{(rand*33).to_i}'"}.join(', ')})")
conn.exec("CREATE TABLE inet6_test (#{(0...cols).map{|i| "i#{i} inet DEFAULT '#{IPAddr.new((rand*inet6_max).to_i, Socket::AF_INET6)}/#{(rand*129).to_i}'"}.join(', ')})")

conn.exec("INSERT INTO inet4_test DEFAULT VALUES")
conn.exec("INSERT INTO inet6_test DEFAULT VALUES")
log_rows.times do
  conn.exec("INSERT INTO inet4_test SELECT * FROM inet4_test")
  conn.exec("INSERT INTO inet6_test SELECT * FROM inet6_test")
end

[4,6].each do |v|
  conn.exec( "SELECT i0 FROM inet#{v}_test LIMIT 1" ) do |res|
    print "Example IPv#{v} value: #{res.getvalue(0, 0).inspect}\n"
  end
end
puts

puts "Basic Tests"
inet_tests = [
  '1.2.3.4',
  '1.2.3.4/0',
  '1.2.3.4/8',
  '1.2.3.4/16',
  '1.2.3.4/32',
  '1.2.3.4/23',
  '1234:3456:5678:789a:9abc:bced:edf0:f012',
  '1234:3456:5678:789a:9abc:bced:edf0:f012/0',
  '1234:3456:5678:789a:9abc:bced:edf0:f012/32',
  '1234:3456:5678:789a:9abc:bced:edf0:f012/64',
  '1234:3456:5678:789a:9abc:bced:edf0:f012/96',
  '1234:3456:5678:789a:9abc:bced:edf0:f012/128',
  '1234:3456:5678:789a:9abc:bced:edf0:f012/123',
]
inet_tests.each do |ip|
  conn.exec("SELECT '#{ip}'::inet") do |res|
    v = res.getvalue(0, 0)
    print "Test inet value: #{ip} should equal #{v}/#{v.prefix}\n"
  end
end

cidr_tests = [
  '0.0.0.0/0',
  '1.0.0.0/8',
  '1.2.0.0/16',
  '1.2.3.4/32',
  '::/0',
  '1234:3456::/32',
  '1234:3456:5678:789a::/64',
  '1234:3456:5678:789a:9abc:bced::/96',
  '1234:3456:5678:789a:9abc:bced:edf0:f012/128',
]
cidr_tests.each do |ip|
  conn.exec("SELECT '#{ip}'::cidr") do |res|
    v = res.getvalue(0, 0)
    print "Test cidr value: #{ip} should equal #{v}/#{v.prefix}\n"
  end
end

conn.exec("SELECT $1::inet, $2::cidr", [IPAddr.new('1.2.3.4'), IPAddr.new('1234:3456:5678:789a:9abc:bced::/96')]) do |res|
  v = res.getvalue(0, 0)
  print "Test inet text encoder values: 1.2.3.4 should equal #{v}/#{v.prefix}\n"
  v = res.getvalue(0, 1)
  print "Test cidr text encoder values: 1234:3456:5678:789a:9abc:bced::/96 should equal #{v}/#{v.prefix}\n"
end

Benchmark.ips do |x|
  x.warmup = -1
  [4,6].each do |v|
    sql = "SELECT * FROM inet#{v}_test"
    x.report("inet#{v}") do
      conn.exec(sql) do |res|
        ntuples = res.ntuples
        recnum = 0
        while recnum < ntuples
          converted_rec = {}
          fieldnum = 0
          while fieldnum < cols
            res.getvalue(recnum, fieldnum)
            fieldnum += 1
          end
          recnum += 1
        end
      end
    end
  end
end

=begin
IPv4
Strings: ~150 ips
Pure Ruby IPAddr: ~10.75 ips
C rb_obj_alloc + rb_ivar_set: ~128 ips
C rb_class_new_instance + mask!: ~66 ips
C rb_class_new_instance + mask: ~58 ips

IPv6
Strings: ~128 ips
Pure Ruby IPAddr: ~7.5 ips
C rb_obj_alloc + rb_ivar_set: ~87 ips
C rb_class_new_instance + mask!: ~43 ips
C rb_class_new_instance + mask: ~40 ips
=end
```


lib/pg/text_encoder.rb
```diff
@@ -1,6 +1,7 @@
 # -*- ruby -*-
 
 require 'json'
+require 'ipaddr'
```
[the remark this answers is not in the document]
jeremyevans: It is not necessary, though it may be useful for documentation purposes.  I can remove the ipaddr require or add one for bigdecimal in the other pull request depending on what the consensus is.

ext/pg_text_decoder.c
```diff
@@ -726,9 +743,127 @@ pg_text_dec_timestamp_without_time_zone(t_pg_coder *conv, char *val, int len, in
...
+		unsigned long long * dstllp = (unsigned long long *)dst;
+
+		if (mask == -1) {
+			mask = 128;
+		} else if (mask < 0 || mask > 128) {
+			rb_raise(rb_eTypeError, "invalid mask for IPv6: %d", mask);
+		}
+		vmasks = s_vmasks6;
+
+		if (!is_big_endian) {
```
larskanis: @jeremyevans What do you think about using `read_nbo128()` [from utils.h](https://github.com/ged/ruby-pg/blob/master/ext/util.h#L44) (yet to be implemented)?
```
unsigned long long dstll = read_nbo128(dst);
ip_int = ULL2NUM(dstll);
```
This would also work if dst is non-aligned on architectures which don't allow non-alligned memory access.

jeremyevans: `unsigned long long` can only hold 64-bit data, not 128-bit, so I don't think that approach is possible. Using `read_nbo64` could significantly simplify the code, though.

PR #26, opened 2018-06-23
WIP: handle UTC times correctly in decoder

SamSaffron:
This still fails specs, but something is not right with handling of timestamp without time zone, stuff is being treated as local time incorrectly. 

I think it may make sense to add a complete integration test here, where we insert into a temp table and then select out of it to avoid needing to do very fancy casting (like in the PR) in tests

### DIFF
--- a/ext/pg_text_decoder.c
+++ b/ext/pg_text_decoder.c
@@ -695,7 +695,7 @@ static VALUE pg_text_decoder_timestamp_do(t_pg_coder *conv, char *val, int len, 
 					}
 					break;
 				}
-				case 1: /* interprete as local time, return as local time */
+				case 1: /* interpret as local time, return as local time */
 					/* Fall through to rb_funcall(), because I couldn't find a fast way to
 					 * use rb_time_timespec_new() for this case.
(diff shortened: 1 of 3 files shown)


SamSaffron:
OK this is ready but since it seemed to be on purpose that we were simply stripping timezone for timestamp without timezone I do not think this default is right for a few reasons. 

1. ActiveRecord and Sequel both default to store UTC times in `timestamp without timezone`. This means that a naive usage of basic type mapper out of the box will cause sadness. 

2. Treating times in the database as a "moving target" where the time you get depends on timezone set locally is very risky and can lead to lots of developer confusion

3. This is faster

On the other hand ... naive casts in PG also simply strip, and there is a way to opt in to local times by making a custom type mapper ... plus this is a breaking change :( 

So I am not sure what the perfect default here is.

SamSaffron:
OK I think it is premature to make this change for 1.1 cause it is too breaking ... maybe we should do this for 2.0 not sure. 

Perhaps the best thing to do here is just to more carefully document it.

PR #27, opened 2018-06-24
Fix various issues in the inet/cidr decoder

jeremyevans:
I found these issues by porting the code to sequel_pg and running
Sequel's tests.

It turns out that IPAddr doesn't support non-zero parts after the
netmask, such as 1.2.3.4/24, and automatically modifies the part
after the netmask to be 0 (turning 1.2.3.4/24 to 1.2.3.0/24).
Because the default is to set instance variables for performance,
this resulted in the creation of IPAddr instances that you could
not generate using IPAddr's public API.  To fix that, mask the
part of the IP address integer(s) before calling UINT2NUM or
ULL2NUM.

I've modified the inet tests to be much simpler, just checking
that the returned value for the inet column matches what you
would get if you parsed the input string with IPAddr.new. I've
also added tests for the above issue.

Modifying the passed in char* buffer is a bad idea, as that
modifies the underlying PG::Result storage directly, leading to
multiple calls to getvalue returning different data.  Instead,
copy the input string to a stack buffer work off of that if
there is a netmask.  This uses a buffer size of 64, but I think
the maximum for IPv6 is only 40 if you count the trailing string
terminator.

This also calls rb_global_variable for s_IPAddr so that you don't
get undefined behavior if the IPAddr constant is garbage collected.

### DIFF
--- a/ext/pg_text_decoder.c
+++ b/ext/pg_text_decoder.c
@@ -827,18 +827,29 @@ pg_text_dec_inet(t_pg_coder *conv, char *val, int len, int tuple, int field, int
 	VALUE ip_int;
 	VALUE vmasks;
 	char dst[16];
+	char buf[64];
 	int af = strchr(val, '.') ? AF_INET : AF_INET6;
 	int mask = -1;
+
+	if (len >= 64) {
+		rb_raise(rb_eTypeError, "too long data for text inet converter in tuple %d field %d", tuple, field);
+	}
 
 	if (len >= 4) {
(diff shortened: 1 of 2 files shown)


larskanis:
Thanks for following up! The contract about read-only input string was [already there](https://github.com/ged/ruby-pg/blob/master/ext/pg_text_decoder.c#L19-L21), but now I enforce it per `const char*`: 19f3bbeafd5902db471b917f5058c3323849a6b4

PR #29, opened 2018-06-29
WIP: Add lazy type casts

larskanis:
This is related to [Project PostgreSQL Hyper-record](https://community.rubybench.org/t/project-postgresql-hyper-record/173).


ext/pg_tuple.c
```diff
@@ -0,0 +1,435 @@
...
+/*
+ * Document-method: allocate
+ *
+ * call-seq:
+ *   PG::VeryTuple.allocate -> obj
+ */
+static VALUE
+pg_tuple_s_allocate( VALUE klass )
+{
+	return TypedData_Wrap_Struct( klass, &pg_tuple_type, NULL );
```
ged: I just learned about this way of writing extensions. Good to have a really clear example of how to do it.

larskanis: Thanks for reviewing Michael! The `Data_Wrap_Struct` way is actually [deprecated in the docs](https://github.com/ruby/ruby/blob/trunk/doc/extension.rdoc#encapsulate-c-data-into-a-ruby-object).

larskanis:
I pondered forth and back how we can implement lazy materialization of values at the best. There were different ideas:
1. Adding a flag (or several) in PG::Result, which enables lazy objects returned by the various array or hash returning or yielding methods.
2. Adding optional arguments to enable lazy objects to some of the array and hash returning methods.
3. Adding new methods which explicit retrieve objects with lazy behaviour.

The main issue with 1 and 2 is that a lazy array or lazy hash is precisely not an ruby Array or Hash. We can always implement only a subset of the functions a real Array/Hash offers. And `{a:1, b:2} == lazy_hash` always fails, because lazy_hash is no real Hash.

Adding lazy materialization to other array or hash returning methods is insofar difficult, that they are fairly different to implement, at least if they shall be fast and memory efficient. And we would have separate objects for hash and array like access.

Therefore this PR is my proposal how I think it's solved at the best. It adds `PG::Result#tuple(idx)` which returns a PG::Tuple which is both a kind of array and kind of hash. And it materializes field values not until accessed. Rails needs hash like access most, but also array like access. This can be handled with just one object.

As en extension to this, I think it makes sense to also add `PG::Result#stream_each_tuple`.

ext/pg_tuple.c
ged: Yeah, I've been meaning to convert all of my C extensions to use the new way, but it helps to see a working example. I just recently found a page called [The Definitive Guide to Ruby's C API](https://silverhammermba.github.io/emberb/c/), which had a bunch of stuff I didn't know or had forgotten as well.

PR #31, opened 2018-08-19
Ignore timeshift for BC timestamp.

larskanis:
Ruby and PostgreSQL seem to have a different opinion about the right timezone to be applied to BC dates in some countries. Therefore verify the timestamp without timezone.
This failed for @ged like so:

1) Basic type mapping PG::BasicTypeMapForResults connection wide type mapping should convert format 0 timestamps with time zone
Failure/Error: expect( res.getvalue(0,2).iso8601(3) ).to eq( Time.new(-4713, 11, 24, 23, 58, 59.1231, "-03:00").getlocal.iso8601(3) )

   expected: "-4713-11-24T18:58:59.123-08:00"
        got: "-4713-11-24T19:06:01.123-07:52"

   (compared using ==)
: # ./spec/pg/basic_type_mapping_spec.rb:261:in `block (5 levels) in <top (required)>'
: # ./spec/helpers.rb:29:in `block in included'


cbandy:
What country was this?

larskanis:
> Would it be better to test all the non-infinity times this way?

That's true. In https://github.com/ged/ruby-pg/pull/31/commits/81ba38f3f51b983f086be618655b6ccd9b4ba9ce I changed all `timestamptz` and added these dates to type_specs so that the timezone conversions are still tested.

ged:
Nice. 👍

PR #32, opened 2018-09-18
Add feature to set testing port.

junaruga:
Hello,
I want to add a feature to set testing port, because in my environment (Fedora project's build environment), I can not use the standard port.

Ref: https://bitbucket.org/ged/ruby-pg/issues/283


larskanis:
Thanks, merged!

PR #33, opened 2018-10-23
Add i386 (32-bit test case) with Docker to Travis CI.

junaruga:
This PR is to add 32 bit case to Travis CI.
It is used for us to investigate 32 bit specific issue, and keep green after fixing the issue.

Here is my result on my repository. I reproduced the issue on Travis.
https://travis-ci.org/junaruga/ruby-pg
https://travis-ci.org/junaruga/ruby-pg/jobs/445214740

The ticket is here.
https://bitbucket.org/ged/ruby-pg/issues/285


junaruga:
semaphoreci is failed. [1]
I assume this is not related to this PR.

[1] https://semaphoreci.com/ged/ruby-pg/branches/pull-request-33/builds/4

PR #36, opened 2019-04-06
Add PG::RecordCoder for en/decoding of Composite Types

larskanis:
This implements proper encoder and decoder for Composite Types as described in https://www.postgresql.org/docs/11/rowtypes.html and as requested in https://bitbucket.org/ged/ruby-pg/issues/258

This adds the following classes:
* PG::RecordCoder < PG::Coder
* PG::RecordEncoder < PG::RecordCoder
* PG::RecordDecoder < PG::RecordCoder
* PG::TextEncoder::Record < PG::RecordEncoder
* PG::TextDecoder::Record < PG::RecordDecoder

I first took `PG::RowCoder` as class name,  due to the `ROW(...)` expression provided by PostgreSQL. However this is changed in the second commit, since "Row" is too generic in my opinion and could lead to confusion with `CopyRow`. So I think "Record" expresses this type of data best. 

Unfortunately the names `PG::CompositeCoder`, `PG::CompositeDecoder` and `PG::CompositeEncoder`are already in use (since pg-0.18.0). They should have been named `PG::ArrayCoder` instead, but now it's probably too late to change them, since they have been released years ago. Actually I doubt that `PG::CompositeCoder` etc. are in use in any external code, but I don't want to break the API for the sake of consistency.

### DIFF
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -26,8 +26,8 @@ ext/pg_coder.c
 ext/pg_connection.c
 ext/pg_copy_coder.c
 ext/pg_errors.c
+ext/pg_record_coder.c
 ext/pg_result.c
-ext/pg_row_coder.c
 ext/pg_text_decoder.c
 ext/pg_text_encoder.c
 ext/pg_tuple.c
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -27,6 +27,7 @@ ext/pg_connection.c
 ext/pg_copy_coder.c
 ext/pg_errors.c
 ext/pg_result.c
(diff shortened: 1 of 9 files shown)


crashtech:
I'm looking forward to this feature!

larskanis:
@ged After almost 3 years of development this is now ready to be reviewed and merged.

I accidentally already pushed this patches into the [bitbucket default branch](https://bitbucket.org/ged/ruby-pg/commits/) and don't know how to remove it from there. If this PR is OK to merge I'll sync github and bitbucket again.

PR #37, opened 2019-05-14
Add a test to check a content of the warning.

junaruga:
This PR is to suppress unnecessary warnings in the testing result.

https://github.com/ged/ruby-pg/pull/35#issue-267538593
> As a default behavior of the testing command, there are messages for the stdout and stderr like this.
This situation makes us harder to find a real warning issue like this issue.

`STDERR` does not work. It is not captured by the RSpec stderr testing method.
But `$stderr` works.

According to rubocop style guide, `$stderr` is recommended rather than `STDERR`.

https://github.com/rubocop-hq/ruby-style-guide

> Use $stdout/$stderr/$stdin instead of STDOUT/STDERR/STDIN. STDOUT/STDERR/STDIN are constants, and while you can actually reassign (possibly to redirect some stream) constants in Ruby, you'll get an interpreter warning if you do so.

### DIFF
--- a/lib/pg/basic_type_mapping.rb
+++ b/lib/pg/basic_type_mapping.rb
@@ -301,7 +301,7 @@ class PG::BasicTypeMapForResults < PG::TypeMapByOid
 			format = result.fformat(field)
 			oid = result.ftype(field)
 			unless @already_warned[format][oid]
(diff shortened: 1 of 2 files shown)


junaruga:
gem-windows case is failed on Travis CI. I think that it's not related to this PR.

larskanis:
Nice!

junaruga:
thanks for merging!

PR #39, opened 2019-05-15
Assign a random testing port to consider a multi tests case on a same host

junaruga:
I like to suggest this feature for testing port.

There is a case of failed test for ruby-pg on a build system in our company.

There are some ruby-pg directories in same host
to build and run test on different arch condition.
Then those directories' tests are executed in parallel.

For the case, `tmp_test_specs/setup.log` shows below error
when starting PostgreSQL from spec/helpers.rb.
Because the tests are using same testing port for each other.

```
waiting for server to start....2019-05-13 16:25:07.565 UTC [2106] LOG:
 could not bind IPv6 address "::1": Address already in use
2019-05-13 16:25:07.565 UTC [2106] HINT:  Is another postmaster
already running on port 54321? If not, wait a few seconds and retry.
```

Assigning random port fixes the situation's issue.
As `rand(1000)` returns between 0 and 999,  the logic `54321 + rand(1000)` to set port number is to set port between 54321 and 55320.

Are you fine for this feature?
I appreciate if it is approved for you.

### DIFF
--- a/spec/helpers.rb
+++ b/spec/helpers.rb
@@ -198,7 +198,7 @@ module PG::TestingHelpers
 		@test_pgdata = TEST_DIRECTORY + 'data'
 		@test_pgdata.mkpath
 
-		ENV['PGPORT'] ||= "54321"
+		ENV['PGPORT'] ||= (54321 + rand(1000)).to_s
 		@port = ENV['PGPORT'].to_i
 		ENV['PGHOST'] = 'localhost'
 		@conninfo = "host=localhost port=#{@port} dbname=test"


pvalena:
CI Error:
```
crypto/async/arch/async_win.o:async_win.c:(.text+0x82): undefined reference to `GetCurrentFiber'
```
is not related to this PR.

pvalena:
How does PG server now the port to start with?

 - `ENV['PGPORT']` is set to your preferred port, in case it's not set already (good). Is this further used to start the daemon

 - `@port` is set from the ENV variable, but used only for `@conninfo` 

 - `@conninfo` is set; uses `@port`, but it's is used only for connection to the PG server.

The server will probably need to be [started](https://github.com/ged/ruby-pg/blob/dfea6bbafe145463f94ef0b704334eedc0720f38/spec/helpers.rb#L217) with `@port` too.

Am I missing something? How does the CI pass with this?

junaruga:
> Am I missing something? How does the CI pass with this?

@pvalena Yes, what yo are missing is`PGPORT` is a reserved environment variable by PostgreSQL.
See https://www.postgresql.org/docs/11/libpq-envars.html
> PGPORT behaves the same as the port connection parameter.

If the variable is not set, the default post number '5432' is used from each PostgreSQL command such as `psql`, `pg_ctrl`, `initdb` and `createdb`.

pvalena:
Good to know. Thanks, I should have look that up (I've underestimated the ENV).

pvalena:
LGTM.

junaruga:
Sorry as I could reproduce random port on the shell script level, and this situation strongly depends on our company's build system, I would close this ticket.

https://stackoverflow.com/questions/42004870/seed-for-random-environment-variable-in-bash
The bash's reserved variable `$RANDOM` generates a number between 0 and 32767.
`$((${RANDOM} % 1000))` returns a number between 0 and 999.

```
$ PGPORT="$((54321 + ${RANDOM} % 1000))" bundle exec rake test
$ cat tmp_test_specs/setup.log
...
waiting for server to start....2019-05-16 13:45:46.452 CEST [19430] LOG:  listening on IPv6 address "::1", port 55215
2019-05-16 13:45:46.452 CEST [19430] LOG:  listening on IPv4 address "127.0.0.1", port 55215
2019-05-16 13:45:46.470 CEST [19430] LOG:  listening on Unix socket "/home/jaruga/git/ruby-pg/tmp_test_specs/.s.PGSQL.55215"
...
```

PR #40, opened 2019-08-21
Update the gem license

sergey-alekseev:
https://github.com/ged/ruby-pg/blob/44300258355e9b221dd53bb27e450ee594e1f57d/LICENSE#L3 mentions "2-clause BSDL"

Just raising the license mismatch. Found this during working with https://github.com/github/licensed.

[ci skip]


larskanis:
Thank you!

PR #43, opened 2019-10-05
Change URLs to point to Github

ged:
Since Bitbucket is shutting down support for Mercurial, and most of the work is done via git nowadays, there’s no reason to keep ruby-pg there any longer.

### DIFF
--- a/README-Windows.rdoc
+++ b/README-Windows.rdoc
@@ -35,8 +35,8 @@ Binary gems for windows can be built on Linux, OS-X and even on Windows
 with the help of docker. This is how regular windows gems are built for
 rubygems.org .
 
-To do this, install boot2docker [on Windows](https://github.com/boot2docker/windows-installer/releases)
-or [on OS X](https://github.com/boot2docker/osx-installer/releases) and make
+To do this, install boot2docker {on Windows}(https://github.com/boot2docker/windows-installer/releases)
+or {on OS X}(https://github.com/boot2docker/osx-installer/releases) and make
 sure it is started. A native Docker installation is best on Linux.
 
 Then run:
@@ -50,7 +50,7 @@ containing binaries for all supported ruby versions.
 
 == Reporting Problems
 
-If you have any problems you can submit them via
-[the project's issue-tracker][bitbucket]. And submit questions, problems, or
+If you have any problems you can submit them via {the project's
+issue-tracker}[https://github.com/ged/ruby-pg/issues]. And submit questions, problems, or
 solutions, so that it can be improved.
 
--- a/README-Windows.rdoc
+++ b/README-Windows.rdoc
@@ -35,8 +35,8 @@ Binary gems for windows can be built on Linux, OS-X and even on Windows
 with the help of docker. This is how regular windows gems are built for
 rubygems.org .
 
-To do this, install boot2docker {on Windows}(https://github.com/boot2docker/windows-installer/releases)
-or {on OS X}(https://github.com/boot2docker/osx-installer/releases) and make
+To do this, install boot2docker {on Windows}[https://github.com/boot2docker/windows-installer/releases]
+or {on OS X}[https://github.com/boot2docker/osx-installer/releases] and make
 sure it is started. A native Docker installation is best on Linux.
 
 Then run:
--- a/README.rdoc
+++ b/README.rdoc
@@ -1,7 +1,6 @@
 = pg
 
-home :: https://bitbucket.org/ged/ruby-pg
-mirror :: https://github.com/ged/ruby-pg
+home :: https://github.com/ged/ruby-pg
 docs :: http://deveiate.org/code/pg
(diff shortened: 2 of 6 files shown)


larskanis:
LGTM

README-Windows.rdoc
```diff
@@ -35,8 +35,8 @@ Binary gems for windows can be built on Linux, OS-X and even on Windows
 with the help of docker. This is how regular windows gems are built for
 rubygems.org .
 
-To do this, install boot2docker [on Windows](https://github.com/boot2docker/windows-installer/releases)
-or [on OS X](https://github.com/boot2docker/osx-installer/releases) and make
+To do this, install boot2docker {on Windows}(https://github.com/boot2docker/windows-installer/releases)
```
cbandy: iiuc, these should be curly then square brackets, e.g. `{text}[url]`

README.rdoc
```diff
@@ -78,7 +77,7 @@ There's also {a Google+ group}[http://goo.gl/TFy1U] and a
 want to chat about something.
 
 If you want to install as a signed gem, the public certs of the gem signers
-can be found in {the `certs` directory}[https://bitbucket.org/ged/ruby-pg/src/tip/certs/]
+can be found in {the `certs` directory}[https://github.com/ged/ruby-pg/tree/master/certs]
```
cbandy: It might be possible to use a [relative link](https://help.github.com/en/articles/about-readmes#relative-links-and-image-paths-in-readme-files) here, `[certs]` or `[certs/]`.

larskanis:
@ged I just got a notification from the mail system that my mail to ged at faeriemud dot org couldn't be delivered. So here is it again:

Hi Michael,

I fully agree with you regarding to moving to Github. I could also live
with Gitlab or Bitbucket, but I never really know what I'm doing when
using Mercurial ;-). Since we already have this mirror on Github, it's
best to just label it the canonical repository now.

Regarding issues migration I did a quick attempt with the following
result: https://github.com/larskanis/importtest/issues
I used this tool with some minor changes:
https://github.com/jeffwidman/bitbucket-issue-migration
All issues have the same number on github, so that references still fit.
If it's OK, I can repeat the import to https://github.com/ged/ruby-pg

--
Regards,
Lars

ged:
Sorry about the email bounce. We've been having mysterious hardware issues on our mailserver.

And yeah, please do import the issues, @larskanis . That looks great! 

I'll fix the issues from your review when I get home tonight @cbandy. Thanks for the review.

README.rdoc
ged: Relative links would only work when rendered on Github, right?

cbandy: Yes, true. The word "directory" made me think of a local checkout.

Since this is about installation, an absolute URL makes sense.

larskanis:
@ged Issue migration to GitHub ged/ruby-pg failes probably due to permissions with 404 like here: https://github.com/jeffwidman/bitbucket-issue-migration/issues/88
Could you please set temporary admin permission for me? Alternatively I can send you the patch and commands for doing the migration on your own.

ged:
You apparently can't grant admin permissions on personal accounts, but I'm happy to run the script myself.

larskanis:
@ged Issue migration should work like so:
```sh
$ git clone https://github.com/larskanis/bitbucket-issue-migration --branch nickname
$ cd bitbucket-issue-migration
$ python3 -m venv ./py3   # As described in the README
$ source ./py3/bin/activate
$ pip3 install -r requirements.pip
```
Now run the first batch of issues:
```sh
python3 migrate.py ged/ruby-pg ged/ruby-pg ged
```
This however stopped for me at issue 34 complaining about the missing tag "Documentation". So I renamed "documentation" to "Documentation" on the github page and restarted with:
```sh
python3 migrate.py ged/ruby-pg ged/ruby-pg ged -f 34
```
Then it did run through all remaining 265 issues.

larskanis:
I just noticed, that we already are at #43 due to the pull requests, that are already on this repository and due to the fact that github uses a common counter for issues and pull requests. So we probably either can't migrate issues 1 to 43 from bitbucket or have to add an offset of 43 to all the issues.

ged:
Ah, right okay. I think that since #43 is 8 years old at this point just omitting them should be fine. It might lead to weirdness with old links, but that's probably inevitable anyway.

ged:
Okay, most of the issues are pulled over. I had a few false starts, so the earliest tickets have mismatched numbers, but everything after #47 or so should be correct.

I'll shut down the Bitbucket project later tonight, then maybe push a release when you merge in the record encoder stuff?

larskanis:
@ged imported issues look good - thanks Michael! So I think the bitbucket project can be shut down.

One remaining issue is that the Appveyor project is still linked to bitbucket instead of GitHub. Can you please fix it and possibly update the badge link in the Readme?

PR #306, opened 2019-11-10
Add ability to use symbols as field names instead of strings

### LINKED ISSUE #278: Symbol hashes
**[Original report](https://bitbucket.org/ged/ruby-pg/issue/278) by Вячеслав Краснов (Bitbucket: [slavestys](https://bitbucket.org/slavestys), GitHub: [slavestys](https://github.com/slavestys)).**

----------------------------------------

Hello! Is there a possibility to make pg return symbol hashes because they work much faster than strings

larskanis:
This adds:
* PG::Result#field_name_type=
* PG::Result#field_name_type
* PG::Result#field_names_as
* PG::Connection#field_name_type=
* PG::Connection#field_name_type

Comparison and lookup is faster on symbols than on strings, since the string hash of symbols is generated at Symbol creation. So retrieval of Symbol based field names is somewhat slower, but this is easily outperformed in frameworks that do a lot of lookups.

There are currently two ways to create hashes in C: dynamic and static symbols. Dynamic symbols are GC'ed and static symbols aren't. I added both ways to this PR. Non-GC'able symbols can have unexpected memory bloat, when field names are dynamically generated. Since both ways showed comparable timings, I think we could remove static symbols again.

Mysql2 implements symbol keys since ages, but only [as static symbols](https://github.com/brianmario/mysql2/blob/master/ext/mysql2/result.c#L156-L158). Maybe the possible memory bloat is a reason why `symbolize_keys` isn't used in ActiveRecord.

This raises ruby version requirement to 2.2, since it depends on GC'able dynamic symbols.

Fixes #278


ext/pg_result.c
```diff
@@ -390,6 +395,27 @@ pgresult_get(VALUE self)
 	return this->pgresult;
 }
 
+static VALUE pg_cstr_to_sym(char *cstr, unsigned int flags, int enc_idx)
```
cbandy: The signature of this function could be more clear that it is specific to field names.

ext/pg_result.c
```diff
@@ -400,12 +426,9 @@ static void pgresult_init_fnames(VALUE self)
...
 
 		for( i=0; i<nfields; i++ ){
-			VALUE fname = rb_tainted_str_new2(PQfname(this->pgresult, i));
-			PG_ENCODING_SET_NOCHECK(fname, this->enc_idx);
-			this->fnames[i] = rb_obj_freeze(fname);
+			char *cfname = PQfname(this->pgresult, i);
+			this->fnames[i] = pg_cstr_to_sym(cfname, this->flags, this->enc_idx);
 			this->nfields = i + 1;
-
-			RB_GC_GUARD(fname);
```
cbandy: I suppose the call to (or return from?) `pg_cstr_to_sym` makes this redundant?

ext/pg_result.c
```diff
@@ -583,24 +606,23 @@ pgresult_nfields(VALUE self)
 
 /*
  * call-seq:
- *    res.fname( index ) -> String
+ *    res.fname( index ) -> String or Symbol
  *
  * Returns the name of the column corresponding to _index_.
```
cbandy: It seems worth mentioning `field_name_type` in these cases where "string or symbol" may be returned.

ext/pg_result.c
```diff
@@ -1436,10 +1464,70 @@ pgresult_stream_each_tuple(VALUE self)
...
 
+/*
+ * call-seq:
+ *    res.field_name_type = Symbol
+ *
+ * Set type of field names specific to this result.
+ * It can be set to one of:
+ * * +:string+ to use String based field names
+ * * +:symbol+ to use Symbol based field names
+ * * +:static_symbol+ to use pinned Symbol (can not be garbage collected) - Don't use this, it will probably removed in future.
```
cbandy: Why deprecate `:static_symbol` the moment it is introduced? Is there a use case for having it?

ext/pg_result.c
```diff
@@ -1309,7 +1335,9 @@ yield_tuple(VALUE self, int ntuples, int nfields)
 	int tuple_num;
 	t_pg_result *this = pgresult_get_this(self);
 	VALUE result = pg_new_result(this->pgresult, this->connection);
+	t_pg_result *newthis = pgresult_get_this(result);
 	UNUSED(nfields);
+	newthis->flags = this->flags;
```
cbandy: I don't grok `pg_new_result`, so this is a naive question: should this copy happen inside `pg_new_result`?

ext/pg_result.c
larskanis: A static function like `pg_cstr_to_sym()` could be inlined by the compiler, so it's not safe to protect the VALUE from being GC'ed. However a call to a non-static function is sufficient to protect it from being GC'ed, so the call to `rb_obj_freeze()` is (and was) already enough.

ext/pg_result.c
larskanis: Shrug! I don't care much about static function names.

ext/pg_result.c
larskanis: Yes, I'll change this.

ext/pg_result.c
larskanis: Hmm, good question! If I think about it a bit, I notice that the initial struct values are copied from `this->connection` in `pg_new_result`, but they should be copied from `this` instead. So currently `typemap=` is broken on `stream_each_tuple`, unless it's already set on the connection. I'll think a bit more about it, how this can be solved best.

ext/pg_result.c
larskanis: `:static_symbol` is the classic way to create symbols from C and it is also used by mysql2 gem. So I wanted to be able compare and benchmark it a bit more. So far I think it is mostly obsolete.

larskanis:
Thank you @cbandy for your valuable feedback! I'll come back with an improved version.

ext/pg_result.c
larskanis: I fixed this issue in commit 6a450fae22ddbfd8a75545265e35a9143396870a on the master branch by introducing a result copy function instead of reusing `pg_new_result`. Thank you for highlighting this!

larskanis:
@ged I would appreciate an opinion from you. Is the API OK for you?

ged:
@larskanis : Yes, this looks good to me. Just to clarify: you're including `static_symbol` so you can benchmark against `mysql2`, but it will likely be removed in favor of just `symbol`?

larskanis:
@ged : No, not to compare against `mysql2`, just to compare the two ways of symbol allocation. I noticed that String allocation is faster, but dynamic vs. static Symbol allocation seems to have mostly the same speed. I still kept both versions, so far, just to have them easily available. And as it turned out, the static symbols have to be used on Truffleruby (and they are dynamic there). I'll remove `:static_symbol` type, if you don't like it. Otherwise it can be removed similar to [conn.guess_result_memsize=](https://github.com/ged/ruby-pg/commit/e37e0b7066f04feb4320a995faee648dc691a71b) when we know that no one needs the alternative way.

What do you think about `conn.field_name_type=` to set a default field_name_type for results?

ged:
Ah, okay. The method name makes sense to me, as does keeping the :static_symbol setting. 👍

larskanis:
@ged The second question was more about adding `field_name_type=` to `PG::Connection` to set the default type per connection?

ged:
Oh, then yes I think that'd be useful. I guess I took its utility for granted. ;)

ext/pg_result.c
```diff
@@ -404,6 +406,29 @@ pgresult_get(VALUE self)
 	return this->pgresult;
 }
 
+static VALUE pg_cstr_to_sym(char *cstr, unsigned int flags, int enc_idx)
+{
+	VALUE fname;
+	if( flags & PG_RESULT_FIELD_NAMES_SYMBOL ){
+#ifndef TRUFFLERUBY
```
cbandy: The code here looks correct for `:symbol` and `:string` on truffle ruby, but it looks like `:static_symbol` will produce strings.

ged:
Excellent work as usual sir! 👍

PR #307, opened 2019-11-21
The taint mechanism will be deprecated in Ruby 2.7

kamipo:
The Ruby core team decided to deprecate the taint mechanism in Ruby 2.7
and will remove that in Ruby 3.

https://bugs.ruby-lang.org/issues/16131
https://github.com/ruby/ruby/pull/2476

In Ruby 2.7, `Object#{taint,untaint,trust,untrust}` and related
functions in the C-API no longer have an effect (all objects are always
considered untainted), and are now warned deprecation message.

https://buildkite.com/rails/rails/builds/65054#14bdbd8c-d935-4b80-8842-c81266a1a34e/6-8

### DIFF
--- a/ext/pg.h
+++ b/ext/pg.h
@@ -308,11 +308,6 @@ char *pg_rb_str_ensure_capa                            _(( VALUE, long, char *, 
 		(curr_ptr) = (end_ptr) = RSTRING_PTR(str) \
 	)
 
-#define PG_RB_TAINTED_STR_NEW( str, curr_ptr, end_ptr ) ( \
-		(str) = rb_tainted_str_new( NULL, 0 ), \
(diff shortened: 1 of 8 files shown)


larskanis:
Thank you @kamipo for providing this patch! I restarted the ruby-head test run and it succeeded in the second attempt. Seems to have been a hiccup on the server side.

I think we can remove the calls to `OBJ_INFECT()` as well, since their only effect is to carry the "taint" and "trust" flags over to another object. Do you agree?

kamipo:
> I think we can remove the calls to `OBJ_INFECT()` as well, since their only effect is to carry the "taint" and "trust" flags over to another object. Do you agree?

Looks so. I missed about that.

simi:
Hello. Thanks for working on this patch to everyone involved. Is there anything I can help with to release this?

PR #308, opened 2019-11-23
Add extended error message functions of PostgreSQL-9.6

larskanis:
I noticed that there are some missing functions and constants regarding error messages.

This adds some new constants and:
* PG::Connection#set_error_context_visibility
* PG::Result#verbose_error_message
* PG::Result#result_verbose_error_message (alias)

### DIFF
--- a/ext/extconf.rb
+++ b/ext/extconf.rb
@@ -73,9 +73,10 @@ end
 have_func 'PQsetSingleRowMode' or
 	abort "Your PostgreSQL is too old. Either install an older version " +
 	      "of this gem or upgrade your database to at least PostgreSQL-9.2."
-have_func 'PQconninfo'
-have_func 'PQsslAttribute'
-have_func 'PQencryptPasswordConn'
+have_func 'PQconninfo' # since PostgreSQL-9.3
+have_func 'PQresultVerboseErrorMessage' # since PostgreSQL-9.6
+have_func 'PQsslAttribute' # since PostgreSQL-10
+have_func 'PQencryptPasswordConn' # since PostgreSQL-10
 have_func 'PQresultMemorySize' # since PostgreSQL-12
 have_func 'timegm'
 have_func 'rb_gc_adjust_memory_usage'
--- a/ext/extconf.rb
+++ b/ext/extconf.rb
@@ -74,12 +74,12 @@ have_func 'PQsetSingleRowMode' or
 	abort "Your PostgreSQL is too old. Either install an older version " +
 	      "of this gem or upgrade your database to at least PostgreSQL-9.2."
 have_func 'PQconninfo' # since PostgreSQL-9.3
+have_func 'PQsslAttribute' # since PostgreSQL-9.5
 have_func 'PQresultVerboseErrorMessage' # since PostgreSQL-9.6
-have_func 'PQsslAttribute' # since PostgreSQL-10
 have_func 'PQencryptPasswordConn' # since PostgreSQL-10
 have_func 'PQresultMemorySize' # since PostgreSQL-12
 have_func 'timegm'
-have_func 'rb_gc_adjust_memory_usage'
+have_func 'rb_gc_adjust_memory_usage' # since ruby-2.4
 
 # unistd.h confilicts with ruby/win32.h when cross compiling for win32 and ruby 1.9.1
 have_header 'unistd.h'
--- a/ext/pg.c
+++ b/ext/pg.c
@@ -422,7 +422,7 @@ Init_pg_ext()
 	rb_define_const(rb_mPGconstants, "CONNECTION_SSL_STARTUP", INT2FIX(CONNECTION_SSL_STARTUP));
 	/* Negotiating environment-driven parameter settings. */
 	rb_define_const(rb_mPGconstants, "CONNECTION_SETENV", INT2FIX(CONNECTION_SETENV));
-	/* Internal state: connect() needed. */
+	/* Internal state - PG.connect() needed. */
 	rb_define_const(rb_mPGconstants, "CONNECTION_NEEDED", INT2FIX(CONNECTION_NEEDED));
 
 	/******     PG::Connection CLASS CONSTANTS: Nonblocking connection polling status     ******/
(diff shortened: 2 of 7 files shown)


ext/pg.c
```diff
@@ -451,13 +451,32 @@ Init_pg_ext()
...
 	rb_define_const(rb_mPGconstants, "PQERRORS_DEFAULT", INT2FIX(PQERRORS_DEFAULT));
-	/* Verbose error verbosity level (#set_error_verbosity) */
+	/* Verbose error verbosity level (#set_error_verbosity).
+	 * The VERBOSE mode includes all available fields. */
 	rb_define_const(rb_mPGconstants, "PQERRORS_VERBOSE", INT2FIX(PQERRORS_VERBOSE));
 
+/* PQERRORS_SQLSTATE was introduced in PG-12 together with PQresultMemorySize() */
+#ifdef HAVE_PQRESULTMEMORYSIZE
+	/* Verbose error verbosity level (#set_error_verbosity).
+	 * The SQLSTATE mode includes only the error severity and the SQLSTATE error code, if one is available (if not, the output is like TERSE mode).
```
cbandy: Do these comments appear in the gem documentation?

L466 is identical to the comment for "VERBOSE mode" while L467 says "like TERSE mode."

ext/extconf.rb
```diff
@@ -73,9 +73,10 @@
 have_func 'PQsetSingleRowMode' or
 	abort "Your PostgreSQL is too old. Either install an older version " +
 	      "of this gem or upgrade your database to at least PostgreSQL-9.2."
-have_func 'PQconninfo'
-have_func 'PQsslAttribute'
-have_func 'PQencryptPasswordConn'
+have_func 'PQconninfo' # since PostgreSQL-9.3
+have_func 'PQresultVerboseErrorMessage' # since PostgreSQL-9.6
+have_func 'PQsslAttribute' # since PostgreSQL-10
+have_func 'PQencryptPasswordConn' # since PostgreSQL-10
```
cbandy: Thank you for tracking down these versions. 🤩

ext/pg_connection.c
```diff
@@ -2625,6 +2625,10 @@ pgconn_get_copy_data(int argc, VALUE *argv, VALUE self )
  * * PQERRORS_TERSE
  * * PQERRORS_DEFAULT
  * * PQERRORS_VERBOSE
+ * * PQERRORS_SQLSTATE
+ *
+ * See also corresponding {libpq function}(https://www.postgresql.org/docs/current/libpq-control.html#LIBPQ-PQSETERRORVERBOSITY).
+
```
cbandy: There is an extra line here.

spec/pg/result_spec.rb
```diff
@@ -307,6 +307,26 @@
...
+		res = @conn.get_result; @conn.get_result
+		expect( res.error_message ).to match(/"xyz"/)
+		expect( res.result_error_message ).to match(/"xyz"/)
+	end
+
+	it "provides a verbose error message", :postgresql_96 do
+		@conn.send_query("SELECT xyz")
+		res = @conn.get_result; @conn.get_result
+		expect( res.verbose_error_message(PG::PQERRORS_TERSE, PG::PQSHOW_CONTEXT_ALWAYS) ).to match(/\A.*\n\z/)
+		expect( res.result_verbose_error_message(PG::PQERRORS_VERBOSE, PG::PQSHOW_CONTEXT_NEVER) ).to match(/\n.*\n/)
```
cbandy: These matches aren't immediately obvious to me. Is the first "ends in newline" and the second "contains two newlines"?

Perhaps a comment about the expectation would be good here.

ext/pg_result.c
```diff
@@ -527,6 +527,32 @@ pgresult_error_message(VALUE self)
 	return ret;
 }
 
+#ifdef HAVE_PQRESULTVERBOSEERRORMESSAGE
```
cbandy: We should mention the PostgreSQL version in rdoc when we conditionally define methods. It looks like our convention is a single line at the end of the comment: `Available since PostgreSQL-9.6`

https://github.com/ged/ruby-pg/blob/8c1b2a44abb2a137a7fe78816f7386fac504fd5f/ext/pg_connection.c#L3373-L3381

ext/pg.c
```diff
@@ -451,13 +451,32 @@ Init_pg_ext()
...
+	 *
+	 * Since PostgreSQL-12.
```
cbandy: For consistency, perhaps this should say `Available since PostgreSQL-12`.

larskanis:
Thank you @cbandy ! I hope I addressed all the issues. While being over it I noticed and fixed a lot of broken documentation in PG::Constants.

ext/pg_connection.c
```diff
@@ -4115,6 +4153,10 @@ init_pg_connection()
 	sym_static_symbol = ID2SYM(rb_intern("static_symbol"));
 
 	rb_cPGconn = rb_define_class_under( rb_mPG, "Connection", rb_cObject );
+/* Help rdoc to known the Constants module */
+#if 0
+	rb_mPGconstants = rb_define_module_under( rb_mPG, "Constants" );
```
cbandy: Interesting. Does rdoc not notice this same call in `ext/pg.c`?

spec/pg/result_spec.rb
```diff
@@ -307,6 +307,28 @@
...
+		res = @conn.get_result; @conn.get_result
+		expect( res.error_message ).to match(/"xyz"/)
+		expect( res.result_error_message ).to match(/"xyz"/)
+	end
+
+	it "provides a verbose error message", :postgresql_96 do
+		@conn.send_query("SELECT xyz")
+		res = @conn.get_result; @conn.get_result
+		# PQERRORS_TERSE should give a single line result
+		expect( res.verbose_error_message(PG::PQERRORS_TERSE, PG::PQSHOW_CONTEXT_ALWAYS) ).to match(/\A.*\n\z/)
```
cbandy: It's not a blocker but I think to assert that there's only one newline (at the end) would look like `/\A[^\n]*\n\z/`.

ext/pg_connection.c
larskanis: Our current documentation [about PG::Connection](https://deveiate.org/code/pg/PG/Connection.html) doesn't resolve the include module but prints "Included Modules rb_mPGconstants" instead. This is fixed by this trick.

spec/pg/result_spec.rb
larskanis: The dot in a Regex without postfix "m" essentially means everything except "\n". So `/\A.*\n\z/` should be the same as `/\A[^\n]*\n\z/m` , I think.

PR #312, opened 2019-12-06
Accept Integer and Float as input to TextEncoder::Numeric

### LINKED ISSUE #310: PG::TextEncoder::Numeric only handles BigDecimal
Lots of backstory to this but I'll keep it down to minimal bug report.

As subject goes PG::TextEncoder::Numeric only handles the BigDecimal class explicitly. This means if trying to encode a Float it falls through as a Float still. Then later on in processing it Errors with TypeError (no implicit conversion of Float into String). 

This was using a type map extracted from the connection itself. Reason Floats are being passed in is because data source is producing Floats. I'm assuming the Numeric TextEncoder should handle other subclasses of Numeric too?

I fixed this with a monkey patch as follows:
```ruby
class PG::TextEncoder::Numeric
  def encode(value)
    return value.to_s('F') if value.is_a?(BigDecimal)
    return value.to_s if value.is_a?(Numeric)
    value
  end
end
```

Happy to submit a PR but unsure what way we want to skin this cat. 

Other obvious variant would be using a case statement like is done for InetAddr.

We can explicitly check for Float and raise NotImplementedError for everything else that is a Numeric, on basis of it might need to be checked / added? An error which can be traced back to here would help, I had to dig around for a bit because the encoder "worked" fine previously, it was just in the extension area that was raising the TypeError but that wasn't visible in the stack trace so it was hard to track down.

Loving the work otherwise!

larskanis:
Current `TextEncoder::Numeric` doesn't handle other classes than `BigDecimal`. That doesn't bother as long as `BasicTypeMapping` is used directly to encode values, since it selects the encoder based on the input type. However if a type map is generated by `PG::TypeMapByOid#build_column_map` or manually it fails on Integer and Float values. This is fixed with this PR.

This moves the conversion of  BigDecimal` to C, so that we can make use of our fast Integer and Float encoders. They avoid additional string allocations, which a ruby conversion would do.

Fixes #310

### DIFF
--- a/ext/pg_text_encoder.c
+++ b/ext/pg_text_encoder.c
@@ -50,6 +50,9 @@
 VALUE rb_mPG_TextEncoder;
 static ID s_id_encode;
 static ID s_id_to_i;
+static ID s_id_to_s;
+static ID s_cBigDecimal;
+static VALUE s_str_F;
 
 static int pg_text_enc_integer(t_pg_coder *this, VALUE value, char *out, VALUE *intermediate, int enc_idx);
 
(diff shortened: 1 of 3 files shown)


larskanis:
I also thought about solving the issue by combining all three numeric encoders into one, since Numeric actually combines features of Integer and Float. This attempt is [here](https://github.com/ged/ruby-pg/compare/master...larskanis:combined-numeric-encoder). However I think it's better to keep them separate as in this PR, since that way we are able to adjust our encoders for differencies I might have overlooked now. Also we would have to adjust some corner cases in our tests.

PR #315, opened 2019-12-15
Add an overview about coders and type maps to README

larskanis:
Ideally this should have the format of a table. Unfortunately rdoc doesn't supoort any kind of tables except in the markdown parser. However even the markdown parser supports tables only through raw HTML and doesn't do any automatic linking to internal classes.

I still think such an overview is very helpful for someone who wants to understand or use the type cast system:

Bildschirmfoto von 2019-12-15 14-27-03

Hope this can be in pg-1.2.0.

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -118,7 +118,8 @@ PG::Coder objects can be used to set up a PG::TypeMap or alternatively
 to convert single values to/from their string representation.
 
 The following PostgreSQL column types are supported by ruby-pg (TE = Text Encoder, TD = Text Decoder, BE = Binary Encoder, BD = Binary Decoder):
-* Integer: {TE}[rdoc-ref:PG::TextEncoder::Integer], {TD}[rdoc-ref:PG::TextDecoder::Integer], {BE2}[rdoc-ref:PG::BinaryEncoder::Int2] {BE4}[rdoc-ref:PG::BinaryEncoder::Int4] {BE8}[rdoc-ref:PG::BinaryEncoder::Int8], {BD}[rdoc-ref:PG::BinaryDecoder::Integer]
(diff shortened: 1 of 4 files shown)


README.rdoc
```diff
@@ -117,6 +117,29 @@ to build composite types by assigning an element encoder/decoder.
 PG::Coder objects can be used to set up a PG::TypeMap or alternatively
 to convert single values to/from their string representation.
 
+The following PostgreSQL column types are supported by ruby-pg (TE = Text Encoder, TD = Text Decoder, BE = Binary Encoder, BD = Binary Decoder):
+* Integer: {TE}[rdoc-ref:PG::TextEncoder::Integer], {TD}[rdoc-ref:PG::TextDecoder::Integer], {BE2}[rdoc-ref:PG::BinaryEncoder::Int2] {BE4}[rdoc-ref:PG::BinaryEncoder::Int4] {BE8}[rdoc-ref:PG::BinaryEncoder::Int8], {BD}[rdoc-ref:PG::BinaryDecoder::Integer]
```
cbandy: Maybe another bullet under integer here for the sized types:

```
* Integer: TE, TD, BD
  * BD: Int2, Int4, Int8
```

Do the classes themselves have documentation about when/why each should be used?

larskanis: Good idea! I'll change this.

larskanis:
Thank you for your prompt review!

> The Timestamp section is organized differently, but that seems to coincide with the class names, too. (They're ... different.)

Yes, timestamps was one of the more difficult types for which I invented this overly complicated type map system instead of some simple 1-to-1 mapping. The other major difficulty was COPY integration, which benefits a lot from the type maps, but was quite a challenge to implement.

PR #321, opened 2020-01-07
Don't use markup in gemspec description & summary

olivierlacan:
image

image

These don't render on https://rubygems.org/gems/pg and the gemspec spec 
specifies they should not include examples or formatting: 
- https://guides.rubygems.org/specification-reference/#description


ged:
The gemspec is a generated file, so changes to it directly aren't going to persist. I am in the middle of updating the build system, however, and this is one of the things I aim to fix.

PR #323, opened 2020-01-10
Recognize symlink failure on old Ruby

cbandy:
`File.symlink()` raises `NotImplementedError` prior to MRI 2.3.

Fixes AppVeyor builds for `ruby_version=22`.


ged:
Looks great, thanks!

PR #330, opened 2020-02-07
Avoid type cast warning in build_coder_maps

amarshall:
These columns are of text-like types (name, regproc) which are not mapped. Since they’re ultimately just treated as text, follow our own warning and cast them in SQL to avoid a Ruby type cast warning.

### DIFF
--- a/lib/pg/basic_type_mapping.rb
+++ b/lib/pg/basic_type_mapping.rb
@@ -118,13 +118,13 @@ module PG::BasicTypeRegistry
 	def build_coder_maps(connection)
 		if supports_ranges?(connection)
 			result = connection.exec <<-SQL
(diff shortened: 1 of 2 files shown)


larskanis:
LoL! That's the better fix! 😆

TSMMark:
This appears to break redshift. It doesn't like the `::text` part. `cannot cast type regproc to character varying`

Here's an example of a commit in another project that removes `::text` casting in order to support redshift. https://github.com/npgsql/npgsql/commit/5c7f26674d4c05d4acc6f097a2abe0580124c422

Monkeypatching this method to revert to the old implementation fixes the issue in our codebase.

larskanis:
@TSMMark Could you please open a new issue for the Redshift incompatibility? I think we can solve it similar to your linked proposal.

TSMMark:
@larskanis done #330 thanks for the quick response

PR #331, opened 2020-02-19
Remove truffleruby from allow_failures

eregon:
* TruffleRuby 20.0.0 should pass the CI fine now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,6 @@ matrix:
         - rake gem:windows:x86-mingw32
   allow_failures:
     - rvm: ruby-head
-    - rvm: truffleruby
   fast_finish: true
 
 before_install:
--- a/spec/pg/connection_spec.rb
+++ b/spec/pg/connection_spec.rb
@@ -566,6 +566,8 @@ describe PG::Connection do
 	end
 
 	it "can receive notices while waiting for NOTIFY without exceeding the timeout" do
+		skip if RUBY_ENGINE == 'truffleruby'
+
 		notices = []
 		@conn.set_notice_processor do |msg|
 			notices << [msg, Time.now]


larskanis:
Thank you @eregon for remembering me about this issue! For now I would like to keep the PR open until
- Truffleruby didn't show any regressions in pg for some months and
- I found a better way to test for [notices while waiting](https://travis-ci.org/ged/ruby-pg/jobs/652518867#L641) that doesn't fail on Truffleruby too often.

eregon:
Agreed the second point needs to be addressed first so the CI passes reliably. That spec might also be flakey on other Rubies.

Not sure about the first point, there won't be a release before 20.1.0 which is [in 3 months](https://www.graalvm.org/docs/release-notes/version-roadmap/) so if TruffleRuby 20.0.0 passes I think we should add it as non-allow_failures, so we notice if something breaks again. If it breaks, we can move it to allow_failures, but if we don't change it I don't think anyone will notice if it starts failing.

eregon:
FWIW, we have regression a test for that memcpy issue, so I doubt that will regress.

eregon:
Could we just skip the notices test on TruffleRuby and then merge this?
I'd rather not accumulate issues without anyone noticing, which sounds likely if it's in allow_failures.

eregon:
I rebased and excluded that spec, better test all specs but that one than failing to see any new failure.

eregon:
CI is green.

eregon:
@larskanis It's been about a month and a half now, would it be OK to merge this PR?
Of course, it's OK to add it back to `allow_failures` if it fails.

I can also update the PR to just skip the flakey spec and keep it in `allow_failures`.
I'll still have a way to monitor it doesn't go red, but it means it might be easier on `pg`'s side to break TruffleRuby unintentionally (e.g., adding a spec that relies on `fork` or so).

cbandy:
I wonder why this test fails? Rather than skip the test we can improve it?

I can see the assertions are of durations, but I can't tell what they are proving. The timestamp given to the notice is too close to when wait_for_notify returns? What does being farther apart indicate?

eregon:
I don't know why the test fails, but I know any test depending on timing is ultimately unreliable from my experience on ruby/spec.
I think it's far more important to run the rest of the tests than wait for this test to be improved, which clearly doesn't seem easy.

I don't know what this test expects nor what the failure means, I did not write it.

@larskanis Maybe you can clarify whether this test is important and what the failure means?
If I or @aardvark179 investigated what's the issue with this test on TruffleRuby, would you merge this PR?

eregon:
@larskanis Thanks for improving the test!

PR #337, opened 2020-03-21
Use native rb_mod_ancestors instead of funcall

fatkodima:


cbandy:
I can see this function exists as far back as MRI 1.8: https://github.com/ruby/ruby/blob/ruby_1_8/class.c#L530-L547

Today:
https://github.com/ruby/ruby/blob/5908914ea2dcb666e099747006e4de262b46ed77/class.c#L1156-L1171

larskanis:
Looks good. Thank you!

PR #339, opened 2020-04-22
Logical Replication Support

waratuman:
I'm currently building a data streaming system and would like to use the [streaming replication protocol](https://www.postgresql.org/docs/12/protocol-replication.html).

The biggest change would be the addition of `PGRES_COPY_BOTH` in the `pgconn_get_last_result` function, marking it as an empty result. Otherwise fetching a result would never return since it would be expecting a result to come in.

The `PG::LogicalReplication` may not be the best spot for this code and am happy to take it out.


larskanis:
I agree to the changes to `pgconn_get_last_result()` , but can you please explain why you want to add `PG::LogicalReplication` to this gem? I think it's better shipped it in a separate gem.

waratuman:
Yeah, I was leaning towards removing it before I submitted the PR. The code was just there since that was my end goal and what I wanted tested.

I just removed it.

Thank you!

larskanis:
I merged this. Thank you!

Issue #343, opened 2020-07-21
We are observing a 10-30% regression in aggregate SQL execution time in v1.2.3 in some cases.


kitsunde:
I haven't devised a test for this yet because I'm not sure what to test exactly, but we monitor our SQL execution times. Specifically, with [prometheus_exporter](https://github.com/discourse/prometheus_exporter) we collect durations of SQL execution on each endpoint by hooking into `PG::Connection` in Rails:

```ruby
MethodProfiler.patch(PG::Connection, [
  :exec, :async_exec, :exec_prepared, :send_query_prepared, :query
], :sql)
```

When we upgraded from 1.2.2 to 1.2.3 we observe the following:

<img width="953" alt="Screenshot 2020-07-21 22 30 12" src="https://user-images.githubusercontent.com/48965/88068303-9b5e2400-cba2-11ea-9ce1-3fbbec121f67.png">

<img width="942" alt="Screenshot 2020-07-21 22 30 04" src="https://user-images.githubusercontent.com/48965/88068294-98fbca00-cba2-11ea-86c8-b8228c42aff6.png">

When we reverted the change the reverse happens:

<img width="997" alt="Screenshot 2020-07-21 22 40 03" src="https://user-images.githubusercontent.com/48965/88068712-25a68800-cba3-11ea-9b04-5e09654e745a.png">

<img width="1009" alt="Screenshot 2020-07-21 22 39 36" src="https://user-images.githubusercontent.com/48965/88068702-22130100-cba3-11ea-89ed-0131c2251a34.png">

These are the execution times of 30-50 queries in each graph. Looking through the changes in 1.2.3 I'm assuming the time isn't actually spent on SQL executing in PG but rather overhead in the library itself.

We've never faced the issue with segmentation faults that the 1.2.3 release is supposed to fix despite being on ruby 2.7.x for months with a fair amount of traffic.

larskanis:
@kitsunde Thank you for the detailed report! And sorry for the long delay!

The new pg version 1.3.0 changes a lot regarding the handling of the socket IO. It would be very interesting how your results are performance-wise!

larskanis:
pg-1.3.0 has a new install switch which should improve the query performance a little bit, but it is not enabled per default:
```
gem inst pg -- --disable-gvl-unlock
```
or through bundler:
```
bundle config build.pg --disable-gvl-unlock
```
[...]

PR #349, opened 2020-08-06
Change all T_DATA objects to typed data

larskanis:
The classic way to allocate T_DATA objects is deprecated since Ruby-2.0. The new way with rb_data_type_t structure has same advantages:

1. Type checks are more specific
2. C-ext can emit memory size information
3. C objects can be made GC.compact friendly

Currently [TruffleRuby fails in CI](https://travis-ci.org/github/larskanis/ruby-pg/jobs/715289488#L635), but the issue is fixed in https://github.com/oracle/truffleruby/commit/f460ce7998c4be937ad298ca3f0daac12d4531e2

This makes ruby-pg GC.compact friendly by reverting https://github.com/ged/ruby-pg/commit/7c1756f953e0c405fc0bf987d046c416e5bb061c but using the compactation callback. This hopefully improves performance as described in #343.

### DIFF
--- a/spec/pg/gc_compact_spec.rb
+++ b/spec/pg/gc_compact_spec.rb
@@ -1,0 +1,103 @@
+# -*- rspec -*-
+# encoding: utf-8
+#
+# Tests to verify correct implementation of compaction callbacks in rb_data_type_t definitions.
+#
+# Compaction callbacks update moved VALUEs.
+# In ruby-2.7 they are invoked only while GC.compact or GC.verify_compaction_references.
+# Ruby constants are usually moved, but local variables are not.
+#
+# Effectiveness of the tests below should be verified by commenting the compact callback out like so:
+#
+#   static const rb_data_type_t pg_tmbo_type = {
+#     "PG::TypeMapByOid",
+#     {
+#       (void (*)(void*))pg_tmbo_mark,
+#       (void (*)(void*))-1,
+#       (size_t (*)(const void *))NULL,
+#   //    pg_compact_callback(pg_tmbo_compact),
+#     },
+#
+# This should result in a segmentation fault aborting the whole process.
+# Therefore the effectiveness of only one test can be verified per rspec run.
+
+require_relative '../helpers'
+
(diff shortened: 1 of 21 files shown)


spec/pg/gc_compact_spec.rb
```diff
@@ -0,0 +1,82 @@
+# -*- rspec -*-
+# encoding: utf-8
+
+require_relative '../helpers'
+
+if GC.respond_to?(:compact)
+	describe "GC.compact" do
```
cbandy, line 7: You may be able to indent less using RSpec's [built-in filtering](https://relishapp.com/rspec/rspec-core/docs/filtering/conditional-filters).

```suggestion
describe "GC.compact", if: GC.respond_to?(:compact) do
```

eregon: The filtering is a bit weird though as it still executes the block.
Probably fine here since all the logic is in `before`/`it`.

spec/pg/gc_compact_spec.rb
```diff
@@ -0,0 +1,85 @@
+# -*- rspec -*-
```
eregon, line 1: This file seems to use tabs, but all other specs use the conventional 2-spaces indent, could you re-indent?

larskanis: No, all ruby-pg source files use tabs indent (for some reason).

spec/pg/gc_compact_spec.rb
larskanis: `describe` blocks are still executed in case of `if: false`. Only the execution of `it` blocks is omitted.

larskanis: I moved all code into before, it and after blocks and made use of the :if filtering. Thanks for the hint!

spec/pg/gc_compact_spec.rb
eregon: Oh, I mixed with FFI, sorry.

PR #360, opened 2020-11-19
Fix README example

jackc:
The example fails because the column names `procpid` and `current_query` are incorrect. The correct names are `pid` and `query`.

See: https://www.postgresql.org/docs/current/monitoring-stats.html#PG-STAT-ACTIVITY-VIEW


ged:
Oops, yeah some of the columns were renamed in the 9.x series and I forgot to update the docs. Thanks for the patch.

Issue #342, opened 2020-07-06
Make pg more Ruby-3.0 fiber scheduler friendly


larskanis:
This was raised by @ioquatix in https://github.com/ffi/ffi/issues/797#issuecomment-654153362 :

In order to support the upcoming Ruby 3.0 fiber scheduler, you need to do only one thing:

When you would call a blocking function that does I/O, instead call `IO#wait_readable` or `IO#wait_writable` respectively.

If you want to support Ruby 2.x, then allow the user to inject a custom IO class, which responds to `.new` and implements `#wait_readable` and `#wait_writable`.

You can check how to do it here:

https://github.com/socketry/db-postgres/blob/6fbdd03dd6b04c57c5b7995b60e3f626c674e609/lib/db/postgres/native/connection.rb#L106-L139

We will be augmenting the C interface with some improvements for invoking `IO#wait_readable/wait_writable` but it hasn't landed yet.

Feature description is here: https://bugs.ruby-lang.org/issues/16786

ioquatix:
Awesome, let me know if I can provide any further help :)

cc @tenderlove

jasl:
I just wrote a PoC

```ruby
class PG::Connection
	module FiberScheduler
		[
			:exec, :query, :exec_params, :prepare,
			:exec_prepared, :describe_portal, :describe_prepared
		].each do |method|
			module_eval <<-CODE, __FILE__, __LINE__ + 1
				def #{method}
					Fiber.scheduler.io_wait(socket_io, IO::WRITABLE, 5)
					super
					Fiber.scheduler.io_wait(socket_obj, IO::READABLE, 5)
				end
			CODE
		end
	end
	include FiberScheduler
end
```

It's dirty, but pass tests (set `evt` as Fiber scheduler in `spec_helpers` then disable `connection_spec` and `connection_sync_spec` temporary)

Any directive or guidance?

A question, I'm stucking on adding `evt` to Gemfile but failed to require it for 10 minutes, then I found I need to declare it in `Rakefile`, any historical reason for using `Hoe` not standard `Gemfile`?

jasl:
@larskanis @ioquatix could you take a look PR #364 ?

I'm done a PoC:

- Add a `PG::Connection.use_fiber=` method to enable fiber scheduler support
  - Require Ruby version > 3.0
  - Require `PG::Connection.async_api = true`
- Change behavior of `PG::Connection.async_api = false`
[...]

PR #364, opened 2020-12-23
Improve Ruby 3.0.0 Fiber scheduler support

jasl:
references #342 

- Change behavior of `PG::Connection.async_api = false`
  - It will raise error if Fiber Scheduler set, because the scheduler leading blocking IO will stucking,
- Update `rake spec`, test both with and without fiber on Ruby 3.0 (ruby-head)

Todo:

- `OS thread support` test cases disabled for now, they will hanging, I don't know how to fix

Help needed:

- I don't know my work is right and covered all cases, although test cases pass

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,6 +13,6 @@ gem "rake-compiler-dock", "~>1.0", :group => [:development, :test]
 gem "hoe-bundler", "~>1.0", :group => [:development, :test]
 gem "rspec", "~>3.5", :group => [:development, :test]
 gem "rdoc", "~>5.1", :group => [:development, :test]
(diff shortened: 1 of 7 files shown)


jasl:
`configure: error: library 'eay32' or 'crypto' is required for OpenSSL`

larskanis:
Thank you @jasl for working on the Fiber scheduler support! I used your PR as a first guide what scheduler support is about and how scheduler support can be implemented. Unfortunately it turned out that were a lot more blocking functions than just the query functions, so that #397 was way more work than I expected. Nevertheless it now adds full fiber scheduler support to ruby-pg.

PR #370, opened 2021-02-19
Avoid casting of OIDs to fix compat with Redshift database

### LINKED ISSUE #369: build_coder_maps incompatible with Redshift
> cannot cast type regproc to character varying

#330 appears to break redshift. It doesn't like the `::text` part. `cannot cast type regproc to character varying`

Here's an example of a commit in another project that removes `::text` casting in order to support redshift. https://github.com/npgsql/npgsql/commit/5c7f26674d4c05d4acc6f097a2abe0580124c422

Monkeypatching this method to revert to the old implementation fixes the issue in our codebase.

---

Original comment https://github.com/ged/ruby-pg/pull/330#issuecomment-782173955

larskanis:
Redshift doesn't support OID these casts, so we do OID lookup through the pg_proc table and add type 'name' as an alias for 'text'.

Also clean up the type map generation, so that only arrays are separated and all other types can be registered.

Fixes #369

### DIFF
--- a/lib/pg/basic_type_mapping.rb
+++ b/lib/pg/basic_type_mapping.rb
@@ -40,31 +40,16 @@ module PG::BasicTypeRegistry
 		def initialize(result, coders_by_name, format, arraycoder)
 			coder_map = {}
 
-			_ranges, nodes = result.partition { |row| row['typinput'] == 'range_in' }
-			leaves, nodes = nodes.partition { |row| row['typelem'].to_i == 0 }
-			arrays, nodes = nodes.partition { |row| row['typinput'] == 'array_in' }
-
-			# populate the enum types
-			_enums, leaves = leaves.partition { |row| row['typinput'] == 'enum_in' }
-			# enums.each do |row|
-			#	coder_map[row['oid'].to_i] = OID::Enum.new
-			# end
+			arrays, nodes = result.partition { |row| row['typinput'] == 'array_in' }
 
 			# populate the base types
-			leaves.find_all { |row| coders_by_name.key?(row['typname']) }.each do |row|
+			nodes.find_all { |row| coders_by_name.key?(row['typname']) }.each do |row|
 				coder = coders_by_name[row['typname']].dup
 				coder.oid = row['oid'].to_i
 				coder.name = row['typname']
 				coder.format = format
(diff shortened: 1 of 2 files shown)


larskanis:
@TSMMark Could you try this PR with Redshift?

TSMMark:
yes will report back after tested

TSMMark:
@larskanis Sorry but my bundler isn't allowing me to install the pg gem using this commit hash ref, so I'm not able to dig into it and test right now

```
gem "pg", git: "https://github.com/larskanis/ruby-pg.git", ref: "365008ef71118da01dc9b3f422c9a81ae62602ad"
```

```
Bundler could not find compatible versions for gem "pg":
  In Gemfile:
    postgres_to_redshift was resolved to 0.1.2, which depends on
      pg (~> 1.2.0)

    redshift_simple_migrator was resolved to 0.2.1, which depends on
      pg (>= 0.18)

    sequel-redshift was resolved to 0.0.2, which depends on
      pg

    sequel_pg was resolved to 1.14.0, which depends on
      pg (>= 0.18.0, != 1.2.0)

Could not find gem 'pg (~> 1.2.0)', which is required by gem 'redshift_simple_migrator', in any of the relevant sources:
  https://github.com/larskanis/ruby-pg.git (at 365008e@365008e)
```

Maybe I'm doing something wrong

larskanis:
The dependency of postgres_to_redshift gem is too strict. I relaxed it in https://github.com/larskanis/postgres_to_redshift/commit/d1079144e632fb09f3e4cf5bd095dceffa97360c

So this file resolves for me:
```ruby
source "https://rubygems.org"

gem "pg", git: "https://github.com/larskanis/ruby-pg.git", ref: "365008ef71118da01dc9b3f422c9a81ae62602ad"
gem "postgres_to_redshift", git: "https://github.com/larskanis/postgres_to_redshift", ref: "d1079144e632fb09f3e4cf5bd095dceffa97360c"
gem "redshift_simple_migrator"
gem "sequel-redshift"
gem "sequel_pg"
```

TSMMark:
That indeed resolves. Should be able to test soon

TSMMark:
LGTM

PR #376, opened 2021-04-08
Improve performance of type map creation

amarshall:
Please see individual commits for extensive details. I’m happy to separate into multiple PRs if some work needs more vetting or discussion, as each commit can stand on its own (but separating will require resolving conflicts as commits 2 & 3 both depend on 1).

This is the result of investigation after encountering some performance issues with type map initialization being unexpectedly slow (after implementing a more time-sensitive use case) when the number of rows queries from `pg_type` is quite large (~131k in production). The enclosed performance testing was done on in an ideal (and faster) development environment with similar schema and slightly large pg_type set (~189k). In our production environment we were seeing type map creation take 4,102±550 ms, and since we create two type maps, that time doubles currently. Needless to say, that is quite slow and these changes aim to improve that.

Finally, I am curious if it is possible to reduce the number of rows which are actually used to generate the coder maps, but I’m not certain if all of there are truly needed. It appears that every table with an array column gets two rows (one array_in, one record_in); since we have many tables this ends up summing to quite a lot.

### DIFF
--- a/lib/pg/basic_type_mapping.rb
+++ b/lib/pg/basic_type_mapping.rb
@@ -87,8 +87,6 @@ module PG::BasicTypeRegistry
 		end
 	end
 
-	private
-
 	def build_coder_maps(connection)
 		result = connection.exec(<<-SQL).to_a
 			SELECT t.oid, t.typname, t.typelem, t.typdelim, ti.proname AS typinput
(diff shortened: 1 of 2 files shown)


amarshall:
@larskanis Any thoughts on this?

PR #378, opened 2021-04-26
Fix some typos

kamipo:


larskanis:
So many typos! I wonder that it was understandable at all. Thank you!

Issue #387, opened 2021-06-24
Multiple hosts are not supported because it's not a valid URI


levkk:
Hello,

Multiple hosts, as documented by [33.1.1.3. Specifying Multiple Hosts](https://www.postgresql.org/docs/12/libpq-connect.html#LIBPQ-CONNSTRING), is not supported because the gem attempts to [parse and re-stringify](https://github.com/ged/ruby-pg/blob/master/lib/pg/connection.rb#L51) the URL before giving it to `PQconnectdb`.

Instead I think it should just pass the connection string as-is to `PQconnectdb` since that function is able to handle all the options and would give an error if the URI is not valid.

Thank you.

larskanis:
I overhauled the connection option builder as part of #397, so that it is possible to specify multiple hosts.

The URI is no longer changed by the option hash, but we still extend it with `hostaddr` and `fallback_application_name` for convenience. I think this is a good tradeoff.

PR #388, opened 2021-06-24
Pass URI directly to PQconnectdb

levkk:
Fixes https://github.com/ged/ruby-pg/issues/387

### DIFF
--- a/lib/pg/connection.rb
+++ b/lib/pg/connection.rb
@@ -49,8 +49,7 @@ class PG::Connection
 		if args.length == 1
 			case args.first
 			when URI, /\A#{URI::ABS_URI_REF}\z/
-				uri = URI(args.first)
-				options.merge!( Hash[URI.decode_www_form( uri.query )] ) if uri.query
+				return args.first
 			when /=/
 				# Option string style
 				option_string = args.first.to_s
--- a/lib/pg/connection.rb
+++ b/lib/pg/connection.rb
@@ -49,7 +49,7 @@ class PG::Connection
 		if args.length == 1
 			case args.first
 			when URI, /\A#{URI::ABS_URI_REF}\z/
-				return args.first
+				uri = args.first
 			when /=/
 				# Option string style
 				option_string = args.first.to_s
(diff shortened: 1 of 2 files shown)


mmontagna (approved):

larskanis:
Unfortunately all tests fail. Could you please fix this?

levkk:
> Unfortunately all tests fail. Could you please fix this?

This is always how I like to start my PRs...tells people I mean business ;) Let me try and fix this mess I made...

levkk:
Local tests pass 👍  I excluded the test that allowed the user to override `user`, `password`, and other URI-specific parameters when a URI and a options hash is passed in because that would force the library to parse the URI and will throw the bad URI error if multiple database hosts are used. This is a slight behavior change, I would expect this to be somewhat breaking to some apps that use this strange feature.

cbandy:
As I recall, URI and Hash are accepted together so the caller can set/override parameters like `sslmode` or `connect_timeout` without knowing the content or format of the connection string. It doesn't look like this behavior was documented outside of tests.

Should the caller have to think differently about `user`, `dbname`, etc?

lib/pg/connection.rb
```diff
@@ -49,8 +49,7 @@ def self::parse_connect_args( *args )
 		if args.length == 1
 			case args.first
 			when URI, /\A#{URI::ABS_URI_REF}\z/
-				uri = URI(args.first)
-				options.merge!( Hash[URI.decode_www_form( uri.query )] ) if uri.query
+				uri = args.first
```
cbandy, line 52: I think it's better to have `uri` either always be `URI` or always be `String`. We don't use the structure of `URI` later on, so converting to `String` seems appropriate here.

```ruby
uri = String(args.first)
```

lib/pg/connection.rb
```diff
@@ -71,13 +70,16 @@ def self::parse_connect_args( *args )
...
-			uri.user     = nil if options[:user]
-			uri.password = nil if options[:password]
-			uri.path     = '' if options[:dbname]
-			uri.query    = URI.encode_www_form( options )
-			return uri.to_s.sub( /^#{uri.scheme}:(?!\/\/)/, "#{uri.scheme}://" )
+			query = URI.encode_www_form( options )
+			if uri.is_a? URI
+				uri = uri.to_s
+			end
+			if uri.include?("?")
```
cbandy, line 77: It looks like `?` does not definitely indicate there is a query. 

```
> URI('postgres://u?n:pw@localhost')              
=> #<URI::Generic postgres://u?n:pw@localhost>

> URI('postgres://un:p?w@localhost')
Traceback (most recent call last):
URI::InvalidURIError (bad URI(is not URI?): "postgres://un:p?w@localhost")
```

levkk: That is interesting...and unfortunate. At the same time, I can't parse the URI that may have two hosts or more because that's definitely not valid. Hmm.

Issue #382, opened 2021-06-14
Expired gem certificate


riffraff:
Hey everyone, 

It seems the certificate associated with the gem has expired 

```
$ gem install pg -P MediumSecurity -v '1.2.3' -- --with-pg-config=/opt/local/lib/postgresql13/bin/pg_config
ERROR:  While executing gem ... (Gem::Security::Exception)
    certificate /CN=ged/DC=FaerieMUD/DC=org not valid after 2020-12-23 20:19:51 UTC
```
I can still install this with security disabled. 

I'm on macOS 11.3 (20E232) with ruby 2.7, if that matters.

larskanis:
@ged Do you want to renew your certificate? Or shall I make future releases with my own certificate?

ged:
I committed my latest cert, but that won't change the cert the last gem was signed by. There isn't really any provision for managing certs in the Rubygems infrastructure, so we'd have to re-release with the new cert.

larskanis:
The gem signing cert is updated in pg-1.3.0.

PR #395, opened 2021-08-18
Update my gem-signing cert

ged:
Refs #382


larskanis (approved):
The certificate is valid for the next 4 months only. Is this intentional?

larskanis:
The problem of expired certificates for code signing is usually solved by using a trusted timestamp service. [Here](https://blog.codeinside.eu/2017/11/30/signing-with-signtool-dont-forget-the-timestamp/) is a description how it works. I use such a timestamp service for signing the RubyInstaller for Windows executable: http://time.certum.pl/

However I don't know if Rubygems is able to verify signatures with trusted timestamps to keep gem signatures valid indefinitely. I didn't notice any such feature.

ged:
Yeah, I wish more infrastructure was set up to make it viable, but this at least is something.

PR #396, opened 2021-08-18
Print some install help if libpq wasn't found

larskanis:
This should help to pick the necessary package without googling.


ged (approved):
Very nice.

PR #397, opened 2021-08-25
WIP: Add full compatibility to Fiber.scheduler of Ruby-3.0

### LINKED ISSUE #342: Make pg more Ruby-3.0 fiber scheduler friendly

larskanis:
The aim of this PR is to make PG::Connection fully Fiber.scheduler compatible. That means that all methods that possibly could block, block through the scheduler `io_wait` callback. Internally we switched to async-methods in pg-1.1 for all `exec` methods. This was a good preparation and half of the work for this PR.

This PR follows the strategy of pg-1.1 and adds more `async_*` methods, renames the blocking methods to `sync_*` and uses `PG.async_api` to switch between them. `async_*` versions will be the default in the future.

Compatibility to ruby < 3.0 is not expected for Fiber.scheduler or Async gem compatibility (but for the rest of the gem). Planned to be merged to pg-1.3.

- [x] Use a triggered interceptor for the TCP connection that acts as a gate to transfer data to the PostgreSQL server and back. This shall detect any functions that don't properly trigger the scheduler.
- [x] Establish connection through the scheduler
- [x] Works on Windows (`IO#readpartial` doesn't call the scheduler -> use `read_nonblock` instead, `PG::Connection#block` implemented in ruby)
- [x] keep external nonblocking API fully compatible (while switching to nonblocking API for all internals in this PR)
- [x] add tests for external nonblocking API to verify compatibility
- [x] `async_send` methods (`async_send_query_prepared` etc.)
- [x] `async_exec` methods (`async_prepare` etc.)
- [x] `async_put_copy_data`, `async_put_copy_end`
- [x] `async_get_copy_data`
- [x] `async_get_result`, `async_get_last_result`
- [x] `discard_results`
- [x] `async_reset` reset the backend connection
- [x] `async_set_client_encoding`
- [x] `async_cancel`
- [x] `PQencryptPasswordConn` can block if no algorithm is given
- [x] `PQping` - impossible to implement with nonblocking IO -> uses thread and releases GVL
- [x] `PQconnectStartParams` can block if option `hostaddr` isn't set
- [x] remove GVL unlocking: since every blocking operation is in ruby space now, we could remove GVL unlocking from libpq-functions. One downside is, that the TcpGateScheduler can not abort a blocking operation in case of a CI failure. -> There's a gem-install-switch now, which allows to disable on quest: `--disable-gvl-unlock`
- [x] compare performance without scheduler -> doesn't show any performance regressions
- [x] move new ruby code to C to avoid performance regressions

Fixes #342

/cc @ioquatix


spec/helpers/tcp_gate_scheduler.rb
```diff
@@ -107,15 +105,19 @@ def write(amount=5)
 							read_str = @internal_io.read_nonblock(len)
 							print_data("write fd:#{@internal_io.fileno}->#{@external_io.fileno}", read_str)
 							@external_io.write(read_str)
+							if until_writeable
+								res = IO.select(nil, [until_writeable], nil, 0)
```
ioquatix, line 109: `IO.select` is not compatible with the fiber scheduler FYI.

larskanis: Thanks, but the timeout is zero here, so that it's always nonblocking. The mechanism is so, that in the write direction only as much data is transferred as necessary to make the observed IO writable again. This should trigger any blocking issues while writing at best.

ioquatix: You could just use `until_writeable.wait_writable(0)`?

spec/helpers/scheduler.rb
```diff
@@ -0,0 +1,241 @@
...
+require 'socket'
+
+begin
+	require 'io/nonblock'
+rescue LoadError
+	# Ignore.
+end
+
+module Helpers
+class Scheduler
```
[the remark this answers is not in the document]
ioquatix: Matz wanted to incorporate Async into Ruby as a default gem but I said no. I want everyone to have equal opportunity to implement the scheduler. Maybe it was a bad choice but I want to encourage other engineers (including yourself) to create their own schedulers.

We could try to make a reference design but I think it would create soft-engineering challenges, including encouraging people to use it in a way it would not be suitable, also ossifying a particular implementation into Ruby core.

spec/helpers/tcp_gate_scheduler.rb
```diff
@@ -0,0 +1,214 @@
...
+#  -------------      ----------------------------------------      --------
+#  | scheduler |      | TCPServer                  TCPSocket |      |      |
+#  |   specs   |----->|  port 5444                  port 5432|----->|Server|
+#  -------------  ^   |                                      |      | port |
+#                 '-------  wait_readable:  <-send data--    |      | 5432 |
+#           observe fd|     wait_writable:  --send data->    |      --------
+#                     ----------------------------------------
+
+module Helpers
+class TcpGateScheduler < Scheduler
```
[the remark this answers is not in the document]
ioquatix: I guess you cannot know for certain that some code does or does not block. But we do have one "metric" for blocking, `rb_nogvl`. But this isn't reliable, even if it can be useful.

https://github.com/ruby/ruby/pull/4779

If we are comfortable with assuming "blocking" = "release GVL", then this will catch those cases.

lib/pg/connection.rb
```diff
@@ -65,7 +94,28 @@ def self::parse_connect_args( *args )
...
+
+			options[:hostaddr] ||= if
+					Fiber.respond_to?(:scheduler) &&
+					Fiber.scheduler &&
+					RUBY_VERSION < '3.1.'
+
+				# Use pure Ruby address resolver to avoid blocking of the scheduler.
+				# `IPSocket.getaddress` isn't fiber aware before ruby-3.1.
+				require "resolv"
+				Resolv.getaddress(host)
```
ioquatix, line 114: This is interesting for compatibility.

ged:
Wow what an awesome improvement! I learned a ton about Ruby 3 just from reading this PR.

We are now way past due for a release; should I put one together this weekend?

ioquatix:
This is amazing well done!

lib/pg/connection.rb
```diff
@@ -25,62 +27,135 @@ class PG::Connection
...
-				uri = URI(args.first)
-				options.merge!( Hash[URI.decode_www_form( uri.query )] ) if uri.query
+			when URI, POSTGRESQL_URI
+				uri = args.first.to_s
+				uri_match = POSTGRESQL_URI.match(uri)
+				if uri_match['query']
+					iopts = URI.decode_www_form(uri_match['query']).to_h.transform_keys(&:to_sym)
+				end
+				# extract "host1,host2" from "host1:5432,host2:5432"
+				iopts[:host] = uri_match['hostports'].split(",").map { |hp| hp.split(":", 2)[0] }.join(",")
```
cbandy, line 99: Does this do the right thing when the URI has an IPv6 address that contains `:`?

larskanis: Thank you for checking this! You're right, just splitting on ":" was too opportunistic. #402 should fix this.

larskanis:
> We are now way past due for a release; should I put one together this weekend?

I would like to add #401 before and we should fix #404 and #398 first.

ged:
Okay, sounds good.

ext/pg_connection.c
```diff
@@ -3904,10 +4006,9 @@ pgconn_set_default_encoding( VALUE self )
 
 	if (( enc = rb_default_internal_encoding() )) {
 		encname = pg_get_rb_encoding_as_pg_encoding( enc );
-		if ( pgconn_set_client_encoding_async(self, encname) != 0 )
+		if ( pgconn_set_client_encoding_async(self, rb_str_new_cstr(encname)) != 0 )
 			rb_warning( "Failed to set the default_internal encoding to %s: '%s'",
 			         encname, PQerrorMessage(conn) );
-		pgconn_set_internal_encoding_index( self );
```
drdrsh, line 3910: @larskanis  Sorry for resurrecting this ancient PR. But was removing `pgconn_set_internal_encoding_index` here intentional?

So this has the side effect of `internal_encoding` not being set correctly in the case the `set_client_encoding` call fails with an error for any reason.

This results in UTF-8 strings being interpreted as SQL_ASCII e.g
```
™™™™™™
becomes
\xE2\x84\xA2\xE2\x84\xA2\xE2\x84\xA2\xE2\x84\xA2\xE2\x84\xA2\xE2\x84\xA2
```
This not the case in version 1.2.3

PR #403, opened 2021-10-03
Add support for pipeline mode of PostgreSQL-14

### LINKED ISSUE #401: Add support for Pipeline mode of PostgreSQL-14
This is a new execution mode of libpq that allows overlapping execution of queries. It is a client side change, that requires libpq of PostgreSQL-14, but all current PostgreSQL servers accept these overlapping queries. It is described here: https://www.postgresql.org/docs/14/libpq-pipeline-mode.html

There are 5 methods to be implemented for pipeline support as well as proper tests of this mode.

This ticket is about basic low level support for the pipeline mode only. Possibly there could be some convenience functions in the future and extensions to other libraries like ActiveRecord.

larskanis:
Fixes #401

### DIFF
--- a/.github/workflows/source-gem.yml
+++ b/.github/workflows/source-gem.yml
@@ -54,8 +54,8 @@ jobs:
             PGVER: "14"
           - os: macos
             ruby: "head"
-            PGVERSION: 14.0-1-osx
-            PGVER: "14"
+            PGVERSION: 13.2-1-osx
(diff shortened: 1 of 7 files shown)


larskanis:
Seems to hang sometimes in pipeline mode -> send_flush_request like so: https://github.com/larskanis/ruby-pg/runs/3784028051?check_suite_focus=true

PR #412, opened 2021-12-16
[RFC] Change clog link on README to relative path

TETRA2000:
I found that the link to clog file on README is broken on GitHub.
It indicates to `https://github_com/***/History_rdoc.html` instead of `https://github.com/***/History_rdoc` .

This can be fixed by changing the URL to [local file links](https://docs.ruby-lang.org/en/2.1.0/RDoc/Markup.html#class-RDoc::Markup-label-Links) on RDoc.
But I'm not sure this is the best way to handle this problem because I don't know how README is used on other places.

I would be grateful if you could give me an advice.
Thank you.

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -2,7 +2,7 @@
 
 home :: https://github.com/ged/ruby-pg
 docs :: http://deveiate.org/code/pg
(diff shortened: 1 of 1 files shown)


larskanis:
Thank you - it seems to work!

PR #417, opened 2022-01-21
Add a faster Windows specific `rb_io_wait()` implementation

### LINKED ISSUE #416: 1.3.0 version slow on Windows
This is my environment:
Rails 6.1.4.4
Ruby 2.7.4
PostgreSQL 12.6
OS Windows 10

PG gem 1.2.3 is working fine for me, but it does not support ruby 3. So I upgraded to the latest version 1.3.0.rc3. It installed without complaints and the rails app runs ok. I noticed some slowness so I compared 1.2.3 against 1.3.0 and the latter is really slow.
For our main page which does nothing rare, standard selects it went from 120ms to 600ms, same results for other pages. This is a 5x performance hit. Maybe 1.3.0 is not yet optimized for windows.
This is currently blocking our upgrade to ruby 3.

larskanis:
It is based on the code that was removed in commit https://github.com/ged/ruby-pg/pull/397/commits/6c885e8ba0b2f0f83d0476c186d69aa88ee7e2c4

Fixes #416

CC: @jirubio

### DIFF
--- a/ext/extconf.rb
+++ b/ext/extconf.rb
@@ -155,6 +155,7 @@ have_func 'rb_io_wait' # since ruby-3.0
 # unistd.h confilicts with ruby/win32.h when cross compiling for win32 and ruby 1.9.1
 have_header 'unistd.h'
 have_header 'inttypes.h'
+have_header('ruby/fiber/scheduler.h') if RUBY_PLATFORM=~/mingw|mswin/
 
 checking_for "C99 variable length arrays" do
 	$defs.push( "-DHAVE_VARIABLE_LENGTH_ARRAYS" ) if try_compile('void test_vla(int l){ int vla[l]; }')
--- a/ext/pg_connection.c
+++ b/ext/pg_connection.c
@@ -2207,8 +2207,110 @@ pgconn_notifies(VALUE self)
 	return hash;
 }
 
+#if defined(_WIN32)
 
-#if defined(HAVE_RB_IO_WAIT)
+/* We use a specialized implementation of rb_io_wait() on Windows.
+ * This is because rb_io_wait() and rb_wait_for_single_fd() are very slow on Windows.
+ */
+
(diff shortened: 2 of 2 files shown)


ariccio:
Nit: both `WSACreateEvent` and `WSACloseEvent` can fail.

I'm basically the only programmer I've ever met who *actually checks these kinds of return codes* when writing low level code, but you'd be shocked at the number of bugs it's found, and the number of times it's saved my butt when debugging mysterious problems. Code I've written like this has even found bugs in WINE.

Perhaps you could `rb_raise` an `rb_eConnectionBad`, an `rb_ePGerror`, or a new and more specific exception type, in that case?

I usually find myself wrapping calls to handle-closing functions like `WSACloseEvent` (or plain old `CloseHandle`) in a little shim that checks the value, where I usually just `std::terminate`/`abort` (since this shouldn't happen unless I missed something), but aborting wouldn't make much sense here. If anything ever goes wrong that causes these to fail and it's not bubbled up to the interpreter, I bet someone will have a bad day :)

If you're compiling with MSVC, you can even annotate the param with `_Frees_ptr_`, (or `_Post_ptr_invalid_`, `__drv_freesMem`) to get static analysis to detect accidental use of closed event handles. You probably don't need this bit, but it's a cool fun fact that I haven't had an excuse to share for a few years now!

larskanis:
@ariccio Thank you for your notes! I totally agree with you, but I judge this patch as a workaround only and would better invest my time into solving the slowness in ruby core. The code here worked [for 11 years](https://github.com/ged/ruby-pg/commit/5bcc6cae66954dfbb5bf7ca598d15c87e6facc1a) now.

larskanis:
@ariccio If you provide a PR I'll merge it.

ariccio:
I appreciate it! But my local environment is not setup right for compiling ruby modules right now. You're probably right that it's no big deal. I'll come back to this when I can find time!

Thanks for maintaining this package.

PR #420, opened 2022-01-31
Fix high CPU usage when waiting with a timeout on Ruby < 3.0

jasoncodes:
This fixes a regression in pg 1.3.0 where the `pg_rb_io_wait` compatibility shim for Ruby 2.x drops fractional seconds. This results in `wait_socket_readable` waiting only whole seconds and busy-looping any remaining fractional seconds for every wait. Waits below 1000 ms will exclusively busy-loop.

This can be reproduced with the following test script:

```ruby
require 'pg'
require 'benchmark'

connection = PG::Connection.new

bm = Benchmark.measure do
  connection.wait_for_notify(1.5)
end

p bm.total.round(3)
```

The output on Ruby 2.7 will show ~0.5 seconds of CPU usage without this fix when it should be ~0.0 seconds. Ruby 3.x is unaffected.


larskanis:
Unbelievable! Who wrote this? Anyway, good catch! Thank you!

PR #426, opened 2022-02-06
Set a date for v1.3.1 release [ci skip]

y-yagi:


larskanis:
Thanks, this is fixed.

PR #434, opened 2022-02-16
CI: use ruby/setup-ruby

MSP-Greg:
Change MSP-Greg/ruby-setup-ruby@win-ucrt-1 to ruby/setup-ruby@v1


larskanis:
Thank you, great work!

PR #455, opened 2022-05-14
Handle client error after all data consumed in #copy_data for output

stevewoodcock:
During #copy_data for output, if an exception is raised after consuming all the
data from get_copy_data, then the subsequent cleanup triggers a "no COPY in
progress" error when it tries to call get_copy_data again, which causes the
original exception to be lost.

Fix by ignoring PG::Error during get_copy_data cleanup.

### DIFF
--- a/lib/pg/connection.rb
+++ b/lib/pg/connection.rb
@@ -281,7 +281,7 @@ class PG::Connection
 				end
 				while get_result
 				end
-				raise
+				raise err
 			else
 				res = get_last_result
 				if !res || res.result_status != PGRES_COMMAND_OK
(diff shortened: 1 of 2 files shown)


larskanis:
Thank you for fixing another corner case of the copy function!

`copy_data` is a function, that has some funny history on my end. At first I stumbled upon various issues with the COPY function and the related client side implementation on my own. Then I noticed that all other implementations in the company and the internet were wrong in some aspect. So I tried to teach anyone else how to implement it right. Then I wondered if there couldn't be a builtin method in ruby-pg that handles the corner cases right. Unfortunately while testing `copy_data` in 2013, I noticed that all my former implementations were wrong in some way as well. Since then I'm glad, that there is a proper helper in ruby-pg and I encourage everyone to use it. Now - almost a decade later - you found another corner case. Very good catch!

stevewoodcock:
You are most welcome, thanks.

PR #456, opened 2022-05-28
Avoid spurious keyword argument warning on Ruby 2.7

jeremyevans:
There is no reason to use **kwargs in either of these methods,
as connect_start doesn't handle keywords (keywords passed to it
will be treated as a positional hash).

Fixes deprecation warning when using Sequel's postgres adapter in
Ruby 2.7.

### DIFF
--- a/lib/pg.rb
+++ b/lib/pg.rb
@@ -65,8 +65,8 @@ module PG
 
 
 	### Convenience alias for PG::Connection.new.
-	def self.connect( *args, **kwargs, &block )
-		Connection.new( *args, **kwargs, &block )
+	def self.connect( *args, &block )
+		Connection.new( *args, &block )
 	end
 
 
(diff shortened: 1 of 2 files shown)


stanhu:
Actually, I'm seeing this warning in 1.4.0 now:

```
DEPRECATION WARNING: /Users/stanhu/.asdf/installs/ruby/2.7.5/lib/ruby/gems/2.7.0/gems/pg-1.4.0/lib/pg.rb:69: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/Users/stanhu/.asdf/installs/ruby/2.7.5/lib/ruby/gems/2.7.0/gems/pg-1.4.0/lib/pg/connection.rb:660: warning: The called method `new' is defined here
```

jeremyevans:
This pull request to fix the issue was correct.  This issue was reintroduced in 6d61d3de0ba0e8dae6d3c7dc783ab282495df8bb.  It seems broken unnecessarily, since the `**kwargs` are not even used by `new`, just passed to `connect_to_hosts`, which only accepts `*args`.  I'll send another pull request to fix this new issue.

PR #463, opened 2022-06-15
Fix transaction leak on early break/return

ejoubaud:
Having a return/break inside the Connection#transaction block would cause the transaction to neither commit nor rollback after its begin, effectively causing the transaction to remain open and leak into the next calls made to the connection and generating warnings: `WARNING:  there is already a transaction in progress`.

That's because `#transaction` used `rescue/else`, and the `else` block won't get executed in case of early break/return.

Here's the problem:

```
irb(main):001:1* def transaction
irb(main):002:1*   p 'pre'
irb(main):003:1*   yield
irb(main):004:1*   p 'post'
irb(main):005:1* rescue
irb(main):006:1*   p 'rescue'
irb(main):007:1* else
irb(main):008:1*   p 'else'
irb(main):009:1* ensure
irb(main):010:1*   p 'ensure'
irb(main):011:0> end
=> :transaction
irb(main):012:1* transaction do
irb(main):013:1*   p 'in'
irb(main):014:1*   break
irb(main):015:0> end
"pre"
"in"
"ensure"
=> nil
```

Using `ensure` instead of `rescue`'s `else` fixes the problem.

### DIFF
--- a/spec/pg/connection_spec.rb
+++ b/spec/pg/connection_spec.rb
@@ -825,11 +825,10 @@ EOT
 			# abort the per-example transaction so we can test our own
 			@conn.exec( 'ROLLBACK' )
 
-			res = nil
 			@conn.exec( "CREATE TABLE pie ( flavor TEXT )" )
 
 			begin
-				res = @conn.transaction do
+				@conn.transaction do
 					@conn.exec( "INSERT INTO pie VALUES ('rhubarb'), ('cherry'), ('schizophrenia')" )
 					# a prior version would neither commit nor rollback when the block included an early break/return
 					break
--- a/spec/pg/connection_spec.rb
+++ b/spec/pg/connection_spec.rb
@@ -821,6 +821,33 @@ EOT
 			end
 		end
 
+		it "commits even if the block includes an early break/return" do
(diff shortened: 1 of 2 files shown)


spec/pg/connection_spec.rb
```diff
@@ -821,6 +821,33 @@ def interrupt_thread(exc=nil)
...
 
+		it "commits even if the block includes an early break/return" do
+			# abort the per-example transaction so we can test our own
+			@conn.exec( 'ROLLBACK' )
+
+			res = nil
+			@conn.exec( "CREATE TABLE pie ( flavor TEXT )" )
+
+			begin
+				res = @conn.transaction do
```
larskanis, line 832: This assignment and the one above are unnecessary. Could you please remove them.

larskanis (changes requested):
Thank you for your investigation and this pull request!

spec/pg/connection_spec.rb
ejoubaud: Good catch thanks, removed in [48bc9b2](https://github.com/ged/ruby-pg/pull/463/commits/48bc9b26f5bb76d53cb2abb7c7d55da1db0c58ba)

PR #464, opened 2022-06-16
Bump VERSION to 1.4.0

larskanis:
As written in the release notes below, the changes since 1.3.5 are mostly bug fixes. However there are two new methods, which are necessary to implement #459 and which also affect the public API. Moreover the mere size of #459 is another reason to jump to 1.4.

@ged, @cbandy Do you think the same?

### DIFF
--- a/History.rdoc
+++ b/History.rdoc
@@ -1,3 +1,26 @@
+== v1.4.0 [YYYY-MM-DD] Lars Kanis <lars@greiz-reinsdorf.de>
+
+Added:
+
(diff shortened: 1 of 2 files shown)


ged:
Yeah makes sense to me.

cbandy:
> Yeah makes sense to me.

Same. 👍🏻

larskanis:
Thanks for commenting! I did some tests with Rails and their test suite and found no regressions. So I'd like to release 1.4.0 this weekend. Any objections?

ged:
Nope. 👍

PR #28, opened 2018-06-29
Use rb_enc_alias

nobu:
`rb_encdb`-prefixed functions are only for internal use.
use `rb_enc_alias` instead.

PR #472, opened 2022-07-30
Use Encoding::BINARY for JOHAB

eregon:
* A dummy encoding does not understand JOHAB anyway.
* Creating an extra encoding leads to significant complexity and slowdowns
  in Ruby implementations: https://bugs.ruby-lang.org/issues/18949

### DIFF
--- a/spec/pg/connection_spec.rb
+++ b/spec/pg/connection_spec.rb
@@ -1807,10 +1807,10 @@ describe PG::Connection do
 				expect( @conn.internal_encoding ).to eq( Encoding::ASCII_8BIT )
 			end
 
-			it "the connection should use JOHAB dummy encoding when it's set to JOHAB" do
+			it "the connection should use the BINARY encoding when it's set to JOHAB" do
 				@conn.set_client_encoding "JOHAB"
 				val = @conn.exec("SELECT chr(x'3391'::int)").values[0][0]
-				expect( val.encoding.name ).to eq( "JOHAB" )
+				expect( val.encoding ).to eq( Encoding::BINARY )
 				expect( val.unpack("H*")[0] ).to eq( "dc65" )
 			end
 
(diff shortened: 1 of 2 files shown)


spec/pg/connection_spec.rb
```diff
@@ -1877,20 +1877,6 @@
...
-					256.times do |eidx|
-						Encoding::UTF_8.replicate("pgtest-#{eidx}")
-					end
-				end
-
-				# Now allocate the JOHAB encoding with an unusual high index
-				@conn.set_client_encoding "JOHAB"
-				val = @conn.exec("SELECT chr(x'3391'::int)").values[0][0]
-				expect( val.encoding.name ).to eq( "JOHAB" )
-			end
```
eregon, line 1892: Note: this test no longer makes sense/can no longer test anything useful since pg does not create encodings dynamically anymore with this PR.

eregon:
@larskanis Could you review this?

larskanis:
IMHO the code for JOHAB is exactly how ruby encodings are intended to be used: It tries to find a corresponding encoding in the running ruby VM, just in case a future ruby version has support for it. And if the search fails, then a placeholder encoding is created, so that the encoding information isn't lost.

That's all OK, but the time of new character encodings is long over. No one invents character encodings these days. If someone would care about JOHAB, he would have implemented support for it in ruby core. So this is dead code.

We already had some discussion about the uselessness of JOHAB encoding in the past  https://github.com/ged/ruby-pg/pull/28#issuecomment-416211417 where we already removed the creation of aliases to the dummy encoding [here](https://github.com/ged/ruby-pg/commit/a33acaf46715af8e8d40c7cc9de0c7880d83f40c). And it was my intention to remove the code entirely in the near future.

eregon:
Great, thank you for merging.

Issue #491, opened 2022-11-11
Segfault using Connection#port


cbandy:
With pg 1.4.4, library_version 130008:

```bash
# happy
$ ruby -r pg -e 'puts PG::Connection.new(user: "postgres")'
#<PG::Connection:0x0000786712e77ad8>

# happy
$ ruby -r pg -e 'puts PG::Connection.new(user: "postgres").port'
5432

# happy
$ ruby -r pg -e 'puts PG::Connection.new(user: "postgres", port: nil)'
#<PG::Connection:0x00007d046978fe20>

# sad
$ ruby -r pg -e 'PG::Connection.new(user: "postgres", port: nil).port'
-e:1: [BUG] Segmentation fault at 0x0000000000000000

# sad
$ ruby -r pg -e 'PG::Connection.new(user: "postgres", port: "").port'
-e:1: [BUG] Segmentation fault at 0x0000000000000000
```

cbandy:
When `port` is empty in a connection string, libpq does not set it from the environment nor the compile-time default. PQport then returns null, and we segfault in atoi.

larskanis:
The fix was merged in #492 and is released as part of pg-1.4.5.

PR #492, opened 2022-11-12
Return the libpq default port when blank in conninfo

cbandy:
This is one potential fix for #491.

Internally, libpq checks for null or empty string stored in pg_conn_host and falls back to the configured default `--with-pgport`. This does the same for our wrapper around [PQport](https://www.postgresql.org/docs/current/libpq-status.html#LIBPQ-PQPORT).

- https://github.com/postgres/postgres/blob/REL_15_0/src/interfaces/libpq/fe-connect.c#L1140-L1146
- https://github.com/postgres/postgres/blob/REL_15_0/src/interfaces/libpq/fe-connect.c#L2385-L2387

```c
	/*
	 * Next, work out the port number corresponding to each host name.
	 *
	 * Note: unlike the above for host names, this could leave the port fields
	 * as null or empty strings.  We will substitute DEF_PGPORT whenever we
	 * read such a port field.
	 */
```

```c
		/* Figure out the port number we're going to use. */
		if (ch->port == NULL || ch->port[0] == '\0')
			thisport = DEF_PGPORT;
```

It has done this for a long time:

- https://github.com/postgres/postgres/blob/REL_10_0/src/interfaces/libpq/fe-connect.c#L1713-L1715
- https://github.com/postgres/postgres/blob/REL9_6_0/src/interfaces/libpq/fe-connect.c#L1419-L1433
- https://github.com/postgres/postgres/blob/REL8_3_0/src/interfaces/libpq/fe-connect.c#L793-L797

### DIFF
--- a/spec/pg/connection_spec.rb
+++ b/spec/pg/connection_spec.rb
@@ -731,6 +731,31 @@ describe PG::Connection do
 		expect( @conn.options ).to eq( "" )
 	end
 
+	it "connects without port and then retrieves the default port" do
+		gate = Helpers::TcpGateSwitcher.new(
+				external_host: 'localhost',
+				external_port: ENV['PGPORT'].to_i,
+				internal_host: "127.0.0.1",
+				internal_port: 5432,
+				debug: ENV['PG_DEBUG']=='1')
+
+		PG.connect(host: "localhost",
+				port: "",
+				dbname: "test") do |conn|
+			expect( conn.port ).to eq( 5432 )
+		end
+
+		PG.connect(hostaddr: "127.0.0.1",
+				port: nil,
+				dbname: "test") do |conn|
+			expect( conn.port ).to eq( 5432 )
+		end
+
+		gate.finish
+	rescue Errno::EADDRINUSE => err
+		skip err.to_s
+	end
+
 	it "can retrieve hostaddr for the established connection", :postgresql_12 do
 		expect( @conn.hostaddr ).to match( /^127\.0\.0\.1$|^::1$/ )
 	end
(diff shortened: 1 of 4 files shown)


ged (commented):
Do you think it's worth adding tests for this?

cbandy:
I tried, though only briefly. The suite is set up to start a local Postgres on a high port, but this code path requires a working connection on the default port.

larskanis:
My first thought was to follow libpq closely and return nil in this special case. But on the other hand it doesn't make much sense to handle this case specially at all. So I wonder why libpq doesn't return a valid port number in any case. It just increases complexity to handle null and empty port on the user side.

So IMHO this PR is OK how it is. I added a test checking the both port return variations.

larskanis:
Should we add `DEF_PGPORT` to PG::Constants?

ged:
@larskanis Yeah, probably. Nice work on the test btw. Those helpers are awesome.

ged (approved):

spec/pg/connection_spec.rb
```diff
@@ -731,6 +731,31 @@
 		expect( @conn.options ).to eq( "" )
 	end
 
+	it "connects without port and then retrieves the default port" do
+		gate = Helpers::TcpGateSwitcher.new(
+				external_host: 'localhost',
+				external_port: ENV['PGPORT'].to_i,
+				internal_host: "127.0.0.1",
+				internal_port: 5432,
```
cbandy, line 739: > Should we add `DEF_PGPORT` to PG::Constants?

That constant can/should be used here, in case someone configured with some other `--with-pgport`. On the other hand, who does that?!

larskanis: OK, done.

PR #502, opened 2023-02-25
Add Japanese translation

gemmaro:
Hello,

This pull request adds Japanese translation of `README.md`, with the following changes:

* Change Japanese readme document format from RDoc to Markdown
* Add translation Rake task, which
  * Record po4a version
  * Translate readme into Japanese

I introduced [po4a](https://po4a.org/) to manage the translation. If the `README.md` is updated, please run `rake translate`. Then the untranslated parts of the `README.ja.md` will be reverted back to English. So no outdated translations will remain.

To future Japanese translators: my translation might be a bit literal and could be improved with the existing PostgreSQL documentation. Feel free to suggest improvements.

Thank you.


larskanis:
I'm not a big fan of translated documentations, since no one updates the translations after the initial effort. So they are almost certainly outdated. With `po4a` the situation is a bit different and I hope it works better. Thank you for introducing me to this tool and of course for the translation itself!

larskanis:
@gemmaro In https://github.com/ged/ruby-pg/commit/feb64b0b4a9a8c566ebead716fa966b2010dfe8f I updated the README. Now I get English text parts as you wrote above when running `rake translate`. But when I run `rake translate` I also get a lot of formatting changes in the po files regarding line wrapping.

Can you update the Japanese README for the latest change and show me how you did it?

PR #504, opened 2023-02-28
Update translation

gemmaro:
This is a continuation from #502.

The formatting changes in PO files are not much of problem; the meaning does not change before and after formatting.
(Historically, the Gettext suite has reformatted the PO files at the 77th column for cosmetics.)

This pull request has 3 commits:

0. Set metadata for po4a config. This prevents the POT file from being updated by default values. (e.g. "Free Software Foundation, Inc." in copyright holder field)
1. Update PO files and generate translation by running `rake translate`. In this case, this marked some entries in the PO file as fuzzy (like the following), and the portions of that entries in the translated document is now in English. Note that a change that only wraps lines is not necessarily fuzzy. This is because, as mentioned above, there is no semantic difference in a change that only wraps a line.

   ```po
   #. type: Plain text
   #: ../README.md:219
   #, fuzzy
   #| msgid "Cleanup extension files, packaging files, test databases:"
   msgid ""
   "Cleanup extension files, packaging files, test databases.  Run this to "
   "change between PostgreSQL versions:"
   msgstr ""
   "拡張ファイル、パッケージファイル、テストデータベースを一掃するには次のように"
   "します。"
   ```

2. (translator's work) Edit PO file manually and generate translation (`rake translate` again). I don't go into depth here, since [the Gettext manual](https://www.gnu.org/software/gettext//manual/html_node/Editing.html) provides some instructions on how to edit the PO file. I found and fixed 5 fuzzy entries and 2 outdated entries.

Even a slight change in the original document will change various parts of the file for translation. This increases the number of differences in version control, but it also makes the translation process much easier.

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -207,8 +207,7 @@ PgはRuby-3.0で導入された`Fiber.scheduler`に完全に対応していま�
 
     $ bundle install
 
-Cleanup extension files, packaging files, test databases.  Run this to
-change between PostgreSQL versions:
+拡張ファイル、パッケージファイル、テストデータベースを一掃するには、このコマンドを走らせてください。PostgreSQLのバージョンも切り替わります。
 
     $ rake clean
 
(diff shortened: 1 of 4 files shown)


larskanis:
@gemmaro Thank you for your description of the process! This helps to maintain the translation.

Issue #418, opened 2022-01-23
Ruby crash when `Fiber.scheduler` is active


larskanis:
Our scheduler tests sometimes crash with a segfault in `vm_search_cc`. There is no appearance of `pg_ext` in the stack trace, so that it looks like a ruby bug rather than a bug in ruby-pg.

```
/home/lars/comcard/ruby-pg/spec/helpers/tcp_gate_scheduler.rb:175: [BUG] Segmentation fault at 0x0000000000000028
ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-linux]

-- Control frame information -----------------------------------------------
c:0005 p:0021 s:0026 e:000021 METHOD /home/lars/comcard/ruby-pg/spec/helpers/tcp_gate_scheduler.rb:175
c:0004 p:0020 s:0018 e:000015 BLOCK  /home/lars/comcard/ruby-pg/spec/helpers/tcp_gate_scheduler.rb:137 [FINISH]
c:0003 p:---- s:0010 e:000009 CFUNC  :loop
c:0002 p:0009 s:0006 e:000005 BLOCK  /home/lars/comcard/ruby-pg/spec/helpers/tcp_gate_scheduler.rb:133 [FINISH]
c:0001 p:---- s:0003 e:000002 (none) [FINISH]

-- Ruby level backtrace information ----------------------------------------
/home/lars/comcard/ruby-pg/spec/helpers/tcp_gate_scheduler.rb:133:in `block in write'
/home/lars/comcard/ruby-pg/spec/helpers/tcp_gate_scheduler.rb:133:in `loop'
/home/lars/comcard/ruby-pg/spec/helpers/tcp_gate_scheduler.rb:137:in `block (2 levels) in write'
/home/lars/comcard/ruby-pg/spec/helpers/tcp_gate_scheduler.rb:175:in `write_fds'

-- Machine register context ------------------------------------------------
 RIP: 0x00007ff9b77f6ad3 RBP: 0x00007ff9b2347fc0 RSP: 0x00007ff9b484b620
 RAX: 0x0000000000000001 RBX: 0x0000000000000000 RCX: 0x00007ff9b2344f50
 RDX: 0x00007ff9b2347fc0 RDI: 0x00003fc100100001 RSI: 0x00007ff9b2347fc0
  R8: 0x00007ff9ac0398e0  R9: 0x00007ff9b24887e0 R10: 0x00005560c6d12f50
[...]

PR #505, opened 2023-03-07
Update to scheduler to avoid segfault.

ioquatix:
Fixes <https://github.com/ged/ruby-pg/issues/418>.

### DIFF
--- a/spec/helpers.rb
+++ b/spec/helpers.rb
@@ -600,10 +600,6 @@ EOT
 		#@conninfo_gate = @conninfo
 	end
 
-	def scheduler_teardown
-		Fiber.set_scheduler(nil)
-	end
-
 	def scheduler_stop
 		if Fiber.scheduler && Fiber.scheduler.respond_to?(:finish)
 			Fiber.scheduler.finish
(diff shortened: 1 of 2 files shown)


ioquatix:
@larskanis can you check this? TY :)

larskanis:
Thank you so much for helping! I'll review it tomorrow. What I wonder about is that ruby shouldn't segfault if I use the scheduler in a wrong way. But the patch to ruby is only about the test scheduler and documentation, which is great and very helpful, but IMHO doesn't fix the problem that ruby core segfaults when misused. Instead it should raise some kind of error.

ioquatix:
The problem is, the scheduler takes a reference to the fiber but does not mark it during GC. It's a tricky problem because marking objects which have been submitted to the kernel is pretty tricky... in this case we keep a separate hash table, but it's not very efficient.

ioquatix:
> What I wonder about is that ruby shouldn't segfault if I use the scheduler in a wrong way

I agree with this but neither does `Thread::Mutex` retain its wait queue. So it's actually the same problem. https://bugs.ruby-lang.org/issues/18818

larskanis:
Thanks for the pointer! That makes the issue much more understandable. Still it's rather unexpected behaviour.

larskanis:
@ioquatix Thank you so much for helping!

PR #507, opened 2023-03-17
Improve error message when query is canceled during streaming

neilvcarvalho:
When streaming a result in single row mode, if this query is aborted - either because it timed out or some other reason -, `ruby-pg` raises the error:

```
PG::InvalidChangeOfResultFields: number of fields changed in single row
mode from X to 0 - this is a sign for intersection with another query
```

This error message is misleading and might confuse people. The number of fields changed to 0 because the result stopped being streamed.

This commit adds a different error message using the same exception class, explaining that the query was canceled or timed out.

### DIFF
--- a/ext/pg_result.c
+++ b/ext/pg_result.c
@@ -1480,6 +1480,9 @@ pgresult_stream_any(VALUE self, int (*yielder)(VALUE, int, int, void*), void* da
 		if( pgresult == NULL )
 			rb_raise( rb_eNoResultError, "no result received - possibly an intersection with another query");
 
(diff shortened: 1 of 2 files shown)


neilvcarvalho:
My initial thought would be to raise the original error from the database, but I couldn't find a way to do that. I'm not very familiar with C code, though. I'd appreciate help doing that.

larskanis:
I opened #510 to fix this issue. It raises the original exception.

neilvcarvalho:
Thank you, @larskanis! That was a good async pairing session 😁 

I'm closing this PR in favor of #510

Issue #509, opened 2023-03-21
Apparently hung connection


choubacha:
I've been working through an issue where we sometimes, generally on the _first_ request to the database lock up at the level of `exec_params`. These hang indefinitely (tens of hours) and the ruby backtrace stops where the clib begins and I'm trying to ascertain how the code is supposed to timeout but I cannot figure it out. It looks like we end up in an infinite loop [here](https://github.com/ged/ruby-pg/blob/master/ext/pg_connection.c#L3103) which calls `wait_socket_readable` without a timeout.

It looks like the code is then reliant on libpq to return a timeout. However, at the very least, `connect_timeout` does not apply here and from my testing `tcp_user_timeout` does not seem to correct this behavior. I'd expect that after some time the connection should timeout if I've configured timeouts so for it to not is a little concerning.

System in use:
Server is postgres 15
We currently use a pgbouncer cluster to proxy requests to postgres
The query that we block on is not in a transaction.
connect_timeout: 15
tcp_user_timeout: 100k
idle_timeout: 0 (however, the connection is checked out of the pool at the time)
checkout_timeout: 60

choubacha:
We rely on the system settings which are the following:

```
$ sysctl net.ipv4 | grep keepalive
net.ipv4.tcp_keepalive_intvl = 60
net.ipv4.tcp_keepalive_probes = 5
net.ipv4.tcp_keepalive_time = 300
```

My understanding is that this should result in a disconnect after ~10 minutes but that's not happening. The tcp_user_timeout should limit this to 100 seconds, but that's also not happening.

mt-clearhaus:
Related to this, I am experiencing the gem not discovering a dead connection despite the use of `keepalives` parameters.

Reproduction using version 1.4.6 and Ruby 2.7:

<details>
<summary>Click to unfold</summary>

```
$ docker run --rm -it debian:bullseye-slim bash
root@2a1ffed1753a:/# apt update
[...]
root@2a1ffed1753a:/# apt install -y --no-install-recommends postgresql ruby-dev bundler postgresql-server-dev-all gcc
[...]
root@2a1ffed1753a:/# service postgresql start
Starting PostgreSQL 13 database server: main.
root@2a1ffed1753a:/# su - postgres
postgres@2a1ffed1753a:~$ psql -c "CREATE USER myuser WITH SUPERUSER PASSWORD 'mypassword';"
CREATE ROLE
[...]

PR #521, opened 2023-04-06
Add PG::Connection#check_connection as a way to verify the socket is OK

larskanis:
... without data exchange with the server.

Related to #509

### DIFF
--- a/lib/pg/connection.rb
+++ b/lib/pg/connection.rb
@@ -343,6 +343,20 @@ class PG::Connection
 		end
 	end
 
+	# Read all pending socket input to internal memory and raise an exception in case of errors.
+	#
+	# This verifies that the connection socket is in a usable state and not aborted in any way.
+	# No communication is done with the server.
+	# Only pending data is read from the socket - the method doesn't wait for any outstanding server answers.
+	#
+	# Raises a kind of PG::Error if there was an error reading the data or if the socket is in a failure state.
+	def check_connection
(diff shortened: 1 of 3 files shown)


lib/pg/connection.rb
```diff
@@ -343,6 +343,22 @@ def ssl_attributes
...
 
+	# Read all pending socket input to internal memory and raise an exception in case of errors.
+	#
+	# This verifies that the connection socket is in a usable state and not aborted in any way.
+	# No communication is done with the server.
+	# Only pending data is read from the socket - the method doesn't wait for any outstanding server answers.
+	#
+	# In order to check the communication with the server, it's recommended to send an empty query per #exec instead of using this method.
+	#
+	# Raises a kind of PG::Error if there was an error reading the data of if the socket is in a failure state.
```
mt-clearhaus, line 354:
```suggestion
	# Raises a kind of PG::Error if there was an error reading the data or if the socket is in a failure state.
```

ged (approved):
Looks good. The suggestion from @mt-clearhaus looks good too.

larskanis:
I wonder if `check_connection` is the best name, as it sounds like a connection round trip is done. Maybe `check_socket` is better. What do you think?

cbandy:
Agreed. I like `check_socket`.

ged:
Yeah, `check_socket` is much better.

PR #525, opened 2023-04-24
Don't overwrite flags of timestamp coders

### LINKED ISSUE #524: Timezone timestamp decoder behavior mismatch with hash vs. keyword arguments on v1.5.0
## Description

The behavior of timestamp convenience classes such as `PG::TextDecoder::TimestampUtc` is different when using a hash vs. keyword arguments.

With a hash argument, we use `Hash#merge`, ensuring that `flags` are always set correctly even if they are passed via the constructor. With keyword arguments, it's possible to accidentally overwrite the value of `flags` such that they are not set correctly (see https://github.com/rails/rails/pull/48055).

```ruby
# Before
class TimestampUtc < Timestamp
  def initialize(params={})
    super(params.merge(flags: PG::Coder::TIMESTAMP_DB_UTC | PG::Coder::TIMESTAMP_APP_UTC))
  end
end

# After
class TimestampUtc < Timestamp
  def initialize(**kwargs)
    super(flags: PG::Coder::TIMESTAMP_DB_UTC | PG::Coder::TIMESTAMP_APP_UTC, **kwargs)
  end
end
```

larskanis:
It is unexpected, since it makes the coder behave contrary to the class name. Moreover it causes a regression in rails:
  https://github.com/rails/rails/issues/48049

Fixes #524

### DIFF
--- a/lib/pg/binary_decoder/timestamp.rb
+++ b/lib/pg/binary_decoder/timestamp.rb
@@ -7,19 +7,19 @@ module PG
 		class TimestampUtc < Timestamp
 			def initialize(hash={}, **kwargs)
 				warn "PG::Coder.new(hash) is deprecated. Please use keyword arguments instead! Called from #{caller.first}" unless hash.empty?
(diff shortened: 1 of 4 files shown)


yahonda:
Thank you for the fix.

Issue #523, opened 2023-04-24
PG::Coder.new(hash) is deprecated


zhugexinxin:
pg 1.5.0
rails 7.0.4.3
<img width="1252" alt="image" src="https://user-images.githubusercontent.com/42615243/233997365-08bcedd2-ef5d-4951-be42-274d88bf8c2f.png">

9mm:
Getting this too

larskanis:
Don't worry about this warning! It is a deprecation warning in pg which has to be fixed in rails. Here is the related rails issue: https://github.com/rails/rails/issues/48046 
The warning will disappear with the next rails patch version.

dogweather:
Maybe this should be logged, rather than a `puts`? (Apologies if that's what's happening.)

a160v:
Hey folks,

The warning message you received is related to passing a hash to the new method of the decoder_class, which is likely a PG::Coder instance. The PG::Coder.new method was updated to use keyword arguments instead of a hash in Ruby 3.0, and the old syntax is now deprecated.

To fix the warning message, you should update the code to use keyword arguments instead of a hash when calling the new method of decoder_class. The issue happens in a line of def _update_typemap_for_default_timezone_:

replace
`@timestamp_decoder = decoder_class.new(@timestamp_decoder.to_h)`
with
`@timestamp_decoder = decoder_class.new(**@timestamp_decoder.to_h)`

By using the double splat operator (**), you can convert the hash returned by @timestamp_decoder.to_h into keyword arguments. This way, the new method of decoder_class receives the arguments in the updated syntax, and the deprecation warning should be resolved.

If you are using a version of Ruby earlier than 3.0, you can continue to use the hash syntax, but be aware that it is now deprecated and may be removed in future versions of Ruby.

larskanis:
pg-1.5.3 marks the warnings as deprecation, so that they are suppressed at an ordinary rails run and shouldn't bother any longer.

dogweather:
> pg-1.5.3 marks the warnings as deprecation, so that they are suppressed at an ordinary rails run and shouldn't bother any longer.

Thank you! This sounds like the right way to go.

PR #528, opened 2023-04-26
Add category to deprecation warnings

wata727:
See https://github.com/ged/ruby-pg/issues/523
See https://ruby-doc.org/3.2.2/Warning.html
See https://bugs.ruby-lang.org/issues/17000

The deprecation warning about `PG::Coder.new(hash)` introduced in v1.5.0 is so frequently reported that it would be more useful if a user had better control over the warning.

This PR sets the category for deprecation warnings. Setting `category: :deprecated` will only show warnings to users who either set `Warning[:deprecated] = true` or set `RUBYOPT=-W:deprecated`. `Warning[:deprecated] = false` by default since Ruby 2.7.2, so it suppresses the warning for most users.

### DIFF
--- a/lib/pg/binary_decoder/timestamp.rb
+++ b/lib/pg/binary_decoder/timestamp.rb
@@ -6,19 +6,19 @@ module PG
 		# Convenience classes for timezone options
 		class TimestampUtc < Timestamp
 			def initialize(hash={}, **kwargs)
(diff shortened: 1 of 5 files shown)


larskanis:
Thanks for the change! Unfortunately it breaks on the current Truffleruby release, but I added #529 to work around it.

Issue #545, opened 2023-10-17
Wrong binary conversion on x86 (32bit intel) arch causing test failures


jackorp:
Trying to package the newest (v1.5.4) ruby-pg gem for Fedora, I am hitting a test failure on 32bit intel arch (x86) related to conversion to binary and back with the `Time` class.

The failing test is `Basic type mapping PG::BasicTypeMapBasedOnResult with usage of result oids for copy encoder selection can type cast #copy_data input with encoder to format 1` here: https://github.com/ged/ruby-pg/blob/v1.5.4/spec/pg/basic_type_map_based_on_result_spec.rb#L121

This happens only with `format 1` which is using the `PG::BinaryEncoder`

The failure:
```
  1) Basic type mapping PG::BasicTypeMapBasedOnResult with usage of result oids for copy encoder selection can type cast #copy_data input with encoder to format 1
     Failure/Error:
       expect( res.values ).to eq( [
       	["\\xff000a0d27", "123", "2023-03-17 03:04:05.678912", "1990-12-17 15:14:45", "12.345", "-1.2345e+168", "2055-12-31", "1234-08-31"],
       	["\\x202078797a2020", "-444", "infinity", "-infinity", "-Infinity", "NaN", "infinity", "-infinity"]
       ] )

       expected: [["\\xff000a0d27", "123", "2023-03-17 03:04:05.678912", "1990-12-17 15:14:45", "12.345", "-1.2345e+168", "2055-12-31", "1234-08-31"], ["\\x202078797a2020", "-444", "infinity", "-infinity", "-Infinity", "NaN", "infinity", "-infinity"]]
            got: [["\\xff000a0d27", "123", "1999-12-31 23:31:17.070656", "1999-12-31 23:50:53.061952", "12.345", "-1.2345e+168", "2055-12-31", "1234-08-31"], ["\\x202078797a2020", "-444", "infinity", "-infinity", "-Infinity", "NaN", "infinity", "-infinity"]

       (compared using ==)
     # ./spec/pg/basic_type_map_based_on_result_spec.rb:137:in `block (5 levels) in <top (required)>'
     # /builddir/build/BUILD/spec/helpers.rb:56:in `block in included'
```

We can see the weird dates are starting with `1999-12-31`, but more importantly only for data created via the `Time` class, notice that the `Date` class does not suffer from this.

Env info:
```
PGSQL version: 15.4
Fedora: Rawhide (F40)
[...]

PR #547, opened 2023-10-20
Explicitly retype timespec fields to int64_t to fix compatibility with 32bit arches.

jackorp:
Timespec fields' time_t type is not guaranteed to be any particular integer. Tests with binary timestamp conversion are failing on 32bit arches (e.g. intel x86) until they are retyped into int64_t, which fixes the issue with encoding the Time instances.

Decoder doesn't need adjusting. It returns the correct time from the encoded binary representation.

Resolves: https://github.com/ged/ruby-pg/issues/545


jackorp:
Note that I have only tested this in Fedora's i386 environment via [mock](https://github.com/rpm-software-management/mock).

But the tests on Fedora with the patch that is included within this PR seems to pass for Fedora supported arches OK: https://koji.fedoraproject.org/koji/taskinfo?taskID=107830936

You can inspect each run (running rspec tests included) by clicking on individual "descendants" buildArch fields and then clicking on the "build.log" that's at the bottom with the "Output" field.

PR #549, opened 2023-10-24
Don't ship POT files in gem.

jackorp:
README.md is already translated when it reaches the gem form, no need to ship the files for translation.


larskanis:
Thanks!

PR #561, opened 2024-03-20
Try to fix crash on truffleruby

larskanis:


larskanis:
This didn't fix the `dead handle` issue, but #563 did.

PR #564, opened 2024-04-08
Make `pgconn_connect_poll` close the socket prior to calling `PQconnectPoll`

jcalvert:
Make `pgconn_connect_poll` close the socket prior to calling `PQconnectPoll`

Since `PQconnectPoll` can change the underlying socket and thus the file descriptor, closing the socket after can manifest a race condition where libpq has closed the socket and another thread has opened a socket with the recycled file descriptor. When the original socket is closed from Ruby, the VM will notify the new thread that the FD has closed (erroneously) resulting in an exception.

This is an exception I encountered in a production environment (C Ruby 3.3), somewhat sporadically  It is quite misleading as the way this manifests is that the exception raises [while waiting](https://github.com/ged/ruby-pg/blob/master/lib/pg/connection.rb#L674) on the socket, so it seems as if connections are getting shared between threads.  

The reason this happens is because when the IO object is closed on the Ruby side, the VM [calls rb_notify_fd_close](https://github.com/ruby/ruby/blob/8191735b73eac4486eebac6530bd92080ee23b9a/io.c#L5661) which in turn sends an exception to any [threads waiting on the file descriptor](https://github.com/ruby/ruby/blob/8191735b73eac4486eebac6530bd92080ee23b9a/thread.c#L2631). On the Ruby 2.x path does not have this problem as `IO.select` will return the exception array, although I believe the VM still calls `close()`

### DIFF
--- a/ext/pg_connection.c
+++ b/ext/pg_connection.c
@@ -515,9 +515,9 @@ static VALUE
 pgconn_connect_poll(VALUE self)
 {
 	PostgresPollingStatusType status;
-	status = gvl_PQconnectPoll(pg_get_pgconn(self));
 
 	pgconn_close_socket_io(self);
+	status = gvl_PQconnectPoll(pg_get_pgconn(self));
 
 	return INT2FIX((int)status);
 }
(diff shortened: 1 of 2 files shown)


larskanis:
Thank you for doing the analysis and the fix!

I changed `PQresetPoll` equally and moved the test to its own `it` example. The test still shows the issue, without running the code of the previous test.

PR #571, opened 2024-07-02
Improved documentation for connection.copy_data on how to use a PG::TypeMapByColumn.

danlo:
Here is how you can use the connection.copy_row type map to insert binary data. 

Understanding PG::TypeMapByColumn is essential, and having an example provides a good starting point for how to convert the data.

I found it pretty hard to find information on example construction for a type map.  Incidentally, ChatGPT completely screws this up.

### DIFF
--- a/Contributors.rdoc
+++ b/Contributors.rdoc
@@ -1,4 +1,3 @@
-
 Thanks to all the great people that have contributed code, suggestions, and patches through the
 years. If you contribute a patch, please include a patch for this file that adds your name to the
 list.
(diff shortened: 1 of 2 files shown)


larskanis:
It isn't difficult to find information about type map usage with `copy_data`. Just click on the particular CopyRow constant and the link will lead you to all the details about possible type maps. There are 4 different CopyRow constants. The binary encoder for instance is here: https://deveiate.org/code/pg/PG/BinaryEncoder/CopyRow.html

PR #581, opened 2024-08-11
Adjust tests with dedicated ssl certificate to fetch it from the test directory

### LINKED ISSUE #576: Test failure with modified `RUBY_PG_TEST_DIR`
I have the source code in deeply nested directory, therefore I am forced to modify `RUBY_PG_TEST_DIR`, because otherwise PG fails to start with `Unix-domain socket path "..." is too long (maximum 107 bytes)` error 🤯. But with the modified `RUBY_PG_TEST_DIR`, I observer these errors:

~~~
Failures:

  1) PG::Connection with async established connection can work with changing IO while connection setup
     Failure/Error: expect( conn.ssl_in_use? ).to be_falsey

       expected: falsey value
            got: true
     # ./spec/pg/connection_spec.rb:571:in `block (3 levels) in <top (required)>'
     # /builddir/build/BUILD/rubygem-pg-1.5.4-build/spec/helpers.rb:56:in `block in included'

  2) running with sync_* methods PG::Connection with async established connection can work with changing IO while connection setup
     Failure/Error: expect( conn.ssl_in_use? ).to be_falsey

       expected: falsey value
            got: true
     # ./spec/pg/connection_spec.rb:571:in `block (4 levels) in <top (required)>'
     # /builddir/build/BUILD/rubygem-pg-1.5.4-build/spec/helpers.rb:56:in `block in included'
~~~

The naive patch to fix this would be:

~~~diff
$ git diff 
diff --git a/spec/pg/connection_spec.rb b/spec/pg/connection_spec.rb
index fcc277b..fd06346 100644
--- a/spec/pg/connection_spec.rb
+++ b/spec/pg/connection_spec.rb
@@ -567,7 +567,7 @@ describe PG::Connection do
                        # Connect with SSL, but use a wrong client cert, so that connection is aborted.
                        # A second connection is then started with a new IO.
                        # And since the pipes above were freed in the concurrent thread above, there is a high chance that it's a lower file descriptor than before.
-                       conn = PG.connect( @conninfo + " sslcert=tmp_test_specs/data/ruby-pg-ca-cert" )
+                       conn = PG.connect( @conninfo + " sslcert=#{TEST_DIRECTORY}/tmp_test_specs/data/ruby-pg-ca-cert" )
                        expect( conn.ssl_in_use? ).to be_falsey
 
                        # The new connection should work even when the file descriptor has changed.
~~~

However, I don't think it is correct. It would likely be better to access this variable:

https://github.com/ged/ruby-pg/blob/629ddb94ebe3c6d1a40f860a8eccfbce40e4c635/spec/helpers.rb#L256

or construct it by different means.

larskanis:
This is important if RUBY_PG_TEST_DIR is set to a different directory.

Fixes #576

### DIFF
--- a/spec/helpers.rb
+++ b/spec/helpers.rb
@@ -247,6 +247,7 @@ module PG::TestingHelpers
 		attr_reader :port
 		attr_reader :conninfo
 		attr_reader :unix_socket
+		attr_reader :pgdata
 
 		### Set up a PostgreSQL database instance for testing.
 		def initialize( name, port: 54321, postgresql_conf: '' )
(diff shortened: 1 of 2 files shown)


cbandy (approved):
LGTM; haven't tried it.

PR #591, opened 2024-09-18
Exclude CI files from built gem

AlexWayfer:
Security inspection of some users or even their clients can be confused by finding out unknown YAML files with `Download` instructions inside final production releases. We can just get rid of them.

Also Regex is fixed now, `/` before closing `)` was breaking.

### DIFF
--- a/pg.gemspec
+++ b/pg.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |spec|
   # The `git ls-files -z` loads the files in the RubyGem that have been added into git.
   spec.files = Dir.chdir(File.expand_path(__dir__)) do
     `git ls-files -z`.split("\x0").reject do |f|
-      f.match(%r{\A(?:test|spec|features|translation|\.github|\.travis|\.appveyor)})
+      f.match(%r{\A(?:test|spec|features|translation|\.)})
     end
   end
   spec.extensions    = ["ext/extconf.rb"]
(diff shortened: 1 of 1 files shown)


pg.gemspec
```diff
@@ -23,7 +23,9 @@ Gem::Specification.new do |spec|
   # Specify which files should be added to the gem when it is released.
   # The `git ls-files -z` loads the files in the RubyGem that have been added into git.
   spec.files = Dir.chdir(File.expand_path(__dir__)) do
-    `git ls-files -z`.split("\x0").reject { |f| f.match(%r{\A(?:test|spec|features|translation)/}) }
+    `git ls-files -z`.split("\x0").reject do |f|
+      f.match(%r{\A(?:test|spec|features|translation|\.github|\.travis|\.appveyor)})
```
cbandy, line 27: 🤔 Perhaps all dot-files (beginning with `.`) should be rejected?

AlexWayfer: Interesting point. I can change to this.

BTW, specifying a list of file is better than "all, but exclude some". Yeah, it can lead to a release error, when you've added new files and forgot to list them in `gemspec`, but it's still more accurate and secure.

But I just wanted to make a small patch instead of discussing an opposite approach.

AlexWayfer: Done. I'm OK with both approaches (list CI files or all dot-files), but, again, in the future it's better to list required files only.

larskanis:
Thank you for the fix!

PR #593, opened 2024-09-18
Add EditorConfig file

AlexWayfer:
It's a common thing which helps to keep unified newlines, indentation, etc. between different developers, and different for different types of files.

More info here: https://editorconfig.org/


larskanis:
Thank you! I never liked the tab indention style, but don't want to change it for easier tracking of changes.

Issue #602, opened 2024-11-13
Feature request: AWS RDS IAM Authentication native support


matt-domsch-sp:
Currently, one must use a password to authenticate to a Postgres server.  AWS RDS's implementation of Postgres allows one to use IAM Authentication, which uses a short-lived generated token in place of the password. 

https://github.com/haines/pg-aws_rds_iam is one attempt to provide this functionality.  This has the drawback that a new short-lived token is generated on every new connection, rather than re-using the short-lived generated token for its 15-minute lifetime.  This can lead to connection failures due to AWS rate limits on generating those tokens.

I'm proposing an incomplete solution here with request for guidance.  This solution retrieves and caches a token, reusing it for the lifetime of the token, then generates another.  Each host/port/user/region combination is cached separately. It's simpler than the above because it replaces the password immediately before establishing the connection by doing so in this gem directly, rather than having to hook all the possible paths a password could be needed via an external gem.

https://github.com/matt-domsch-sp/ruby-pg/commit/ef73bf7d2b2469d83e82aeea4bbe055ec81bc27d

One significant drawback to this approach is that I have not found a good way to add connection options that are only used by the gem, and not used or recognized by libpg, which does option validation.  Here I've avoided that by using environment variables, but that solution isn't optimal. I'd prefer to be able to set `:use_iam_authentication` (to invoke the functionality) and `:host_region` (a value required to generate the token) in the connection configuration iopts, and simply have them be ignored by methods that do not need to know about them.  Advise welcome please.

larskanis:
Sorry for not answering earlier, but I don't want an implementation like #605 in ruby-pg. If it's implemented in libpq, then it's no problem - then it's not in my responsibility to maintain and check it. But in ruby-pg we would need proper tests that work both locally and in the CI without a lot of setup.

Architectural I also think that https://github.com/haines/pg-aws_rds_iam is the better approach. We can talk about easier hooks in ruby-pg, in order to plug-in IAM as authentication method for ruby-pg. But it should be in it's own repository.

matt-domsch-sp:
Feature request for pg-aws_rds_iam filed. https://github.com/haines/pg-aws_rds_iam/issues/688

PR #605, opened 2024-11-20
Initial pass adding AWS IAM Authentication

matt-domsch-sp:
This adds AWS IAM Authentication, which replaces the `password` parameter normally provided with a dynamically generated, short-lived authetication token. The authentication token (password) will be fetched from IAM and cached for the next 14 minutes (tokens expire in 15 minutes).  These can then be reused by all new connections until it expires, at which point a new token will be fetched when next needed.

To allow for multiple configurations to multiple servers, the cache is keyed by database username, host name, port, and region.

To use, specify in the environment variables:
- PG_USE_IAM_AUTHENTICATION=true
- AWS_REGION=us-east-1  or whatever your AWS region is

As prerequisites, you must enable IAM authentication on the RDS instance, create an IAM policy, attach the policy to the target IAM user or role, create the database user set to use the AWS Authentication Plugin, and then run your ruby code using that user or role.  See
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.Connecting.html for details on these steps.

You must include the aws-sdk-rds gem in your bundle to use this feature.  A LoadError is raised if the functionality is tried to be used but that gem is not present.

The use of environment variables to invoke this functionality is unusual. You'd expect to able to put these on the postgres connection string.  However, only arguments valid to PostgreSQL itself are allowed on the connection string, and only this gem needs them.  If there's a better way, please advise.

Implements code in issue https://github.com/ged/ruby-pg/issues/602

### DIFF
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -43,6 +43,7 @@ ext/vc/pg.sln
 ext/vc/pg_18/pg.vcproj
 ext/vc/pg_19/pg_19.vcproj
 lib/pg.rb
+lib/pg/aws_iam_auth.rb
 lib/pg/basic_type_mapping.rb
 lib/pg/binary_decoder.rb
 lib/pg/coder.rb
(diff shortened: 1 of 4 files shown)


matt-domsch-sp:
Closing per comments in #602 .

Issue #618, opened 2024-12-20
async connection API does not properly handle timeouts with multiple hosts


ajmaidak:
If 10.0.0.1 does not respond via TCP even if 10.0.0.2 is online and functioning the following connection will timeout
```
require 'pg'
conn = PG.connect(
  host: '10.0.0.1,10.0.0.2',
  port: 5432,
  dbname: 'postgres',
  user: 'user',
  password: 'password',
  connect_timeout: 5,
  target_session_attrs: 'read-write',
)

result = conn.exec('SELECT version()')
puts result[0]['version']
```

The issue is that the async api supported by libpq ignores the "connect_timeout" connection parameter when using PQconnectStart/PQconnectPoll:

[_The connect_timeout connection parameter is ignored when using PQconnectPoll; it is the application's responsibility to decide whether an excessive amount of time has elapsed._](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PQCONNECTSTARTPARAMS)

[You can see discussion of this on the PostgreSQL hackers mailing list](https://www.postgresql.org/message-id/CA%2Bmi_8YyGKA9dWELu63e%3DKL2oN-%2BFe4uca4EtFfb6uQD4Up8pw%40mail.gmail.com).  Specifically around this same issue with psycopg3. 

Browsing the code a bit I believe when the connect_timeout expires [here](https://github.com/ged/ruby-pg/blob/c6d1047b419e87e57aaec076d9338e80f9bd269a/lib/pg/connection.rb#L691-L712) rather then failing there needs to be a call to [PQreset or PQresetStart/PQresetPoll](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PQRESETSTART) to force libpq to reset the connection and try to access the next host?

larskanis:
pg-1.5.x versions should do proper connection timeout handling. The details are described in this commit: https://github.com/ged/ruby-pg/commit/40b2ad5d8164eea4fb14d0275b6825aaf2d1a562

In your case it should wait 5 seconds per connection and abort after 10 seconds if no connection can be established.

Can you describe how it behaves in your setup? Can you do a bit debugging in the code path of pg that you referenced, to investigate how it behaves exactly?

ajmaidak:
> In your case it should wait 5 seconds per connection and abort after 10 seconds if no connection can be established.

Its not, here an example: 

Run a postgres docker:

```
docker run -d \
    -p 5432:5432 \
    --name some-postgres \
[...]

PR #619, opened 2024-12-23
try the next host on connnect_timeout

ajmaidak:
closes: #618 

libpq does not support connect_timeout when using the async api.

So when using the async connection API  with a multi-host connection string libpq will never timeout connections to the first host in the list and will not try try subsequent hosts in the connection string list.

This fixes this by closing and reopenning the connection with a reordered connection string if the connection to a host times out.

See discussion on the pgsql-hackers list discussing this "feature" of the api: https://www.postgresql.org/message-id/flat/CA%2Bmi_8YyGKA9dWELu63e%3DKL2oN-%2BFe4uca4EtFfb6uQD4Up8pw%40mail.gmail.com

### DIFF
--- a/lib/pg/connection.rb
+++ b/lib/pg/connection.rb
@@ -680,6 +680,7 @@ class PG::Connection
 				host_count = conninfo_hash[:host].to_s.count(",") + 1
 				stop_time = timeo * host_count + Process.clock_gettime(Process::CLOCK_MONOTONIC)
 			end
+			connection_attempts = 1
 
 			poll_status = PG::PGRES_POLLING_WRITING
 			until poll_status == PG::PGRES_POLLING_OK ||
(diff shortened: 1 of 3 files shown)


ajmaidak:
@larskanis any feedback?

larskanis:
Thank very much you for this PR! Somehow I had the opinion, that this is already implemented. But it is not!

I'll try to add some useful tests for it and will probably add an error message that mentions multiple host connection attempts, like libpq does in blocking mode:
```sh
$ psql "host=11.11.11.11,22.22.22.22 connect_timeout=5"
psql: error: connection to server at "11.11.11.11", port 5432 failed: timeout expired
connection to server at "22.22.22.22", port 5432 failed: timeout expired
```

PR #626, opened 2025-01-31
Add cross compiler for aarch64-mingw-ucrt

larskanis:
This adds rake target `rake gem:native:aarch64-mingw-ucrt`.

It needs rake-compiler-dock with support for Windows on Arm64, introduced in: https://github.com/rake-compiler/rake-compiler-dock/pull/152

Patches submitted:
* Postgresql: https://commitfest.postgresql.org/patch/5610/
* OpenSSL: https://github.com/openssl/openssl/pull/26605

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -44,6 +44,7 @@ end
 
 CrossLibrary = Struct.new :platform, :openssl_config, :toolchain
 CrossLibraries = [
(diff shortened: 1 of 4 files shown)


cbandy:
The patch may have been reverted recently.

PR #632, opened 2025-03-08
Change #server_version to raise instead of return 0 on error

byroot:
Ref: https://github.com/rails/rails/issues/54712

Returning 0 on error is a bit uncommon in Ruby, and not very convenient, I think it would make more sense to raise the appropriate exception.

Now I understand that this is a bit of a breaking change, so I'd understand why you wouldn't want to merge this.

cc @joshuay03

### DIFF
--- a/ext/pg_connection.c
+++ b/ext/pg_connection.c
@@ -845,14 +845,23 @@ pgconn_parameter_status(VALUE self, VALUE param_name)
  * call-seq:
  *   conn.protocol_version -> Integer
  *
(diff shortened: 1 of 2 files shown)


larskanis:
This makes sense. I added some more corresponding changes and tests, to make it more complete.

The error case is uncommon in these functions and these function are probably not in heavy use. I therefore think there aren't real world compatibility issues with this breaking change.

Thank you for the patch!

PR #653, opened 2025-07-31
Serialize CompositeCoder#dimensions only when set

### LINKED ISSUE #652: Newly added dimension field makes deserialization of 1.6.0 array field on 1.5.9 raise
Ref: https://github.com/rails/rails/issues/55423

The deserialization code blindy tries to put each key-value pair into a writer method: https://github.com/ged/ruby-pg/blob/b77e86c478c294fd74ed187b7573df2665903209/lib/pg/coder.rb#L17-L23

Since older versions don't have the dimension, trying to assign to it fails. I wonder if the code should be more careful with that and rescue `NameError` or similar. Would that make sense?

This is in the context of a rails app where rolling deploys are happening. Apps with the new code cache the new dimension field and older apps that haven't been rotated yet will fail to read it. Rails actually discourages caching full models for similar reasons but people tend to do so anyways since it works 99.99% of the time.

larskanis:
This fixes the compatibility to pg-1.5.9, when deserializing Marshal data from pg-1.6, as long as the new attribute isn't used.

Fixes #652

### DIFF
--- a/lib/pg/coder.rb
+++ b/lib/pg/coder.rb
@@ -72,12 +72,13 @@ module PG
 
 	class CompositeCoder < Coder
 		def to_h
-			{ **super,
+			h = { **super,
 				elements_type: elements_type,
 				needs_quotation: needs_quotation?,
 				delimiter: delimiter,
-				dimensions: dimensions,
 			}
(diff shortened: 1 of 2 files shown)


Earlopain:
Thanks, this patch makes sense to me! Should help with the majority of cases. I tested it out, works as expected.

Issue #661, opened 2025-08-04
LoadError: cannot load such file -- pg_ext


fabio:
Got this error with Ruby 3.3.4 (`ruby 3.3.4 (2024-07-09 revision be1089c8ec) [x86_64-linux]`) on Debian 10.13 (`Linux 4.19.0-25-amd64 #1 SMP Debian 4.19.289-2 (2023-08-08) x86_64 GNU/Linux`) and **pg '~>1.6'**.  It is OK with **pg 1.5.9**

```
$ rbenv exec bundle info pg
  * pg (1.6.1)
	Summary: Pg is the Ruby interface to the PostgreSQL RDBMS
	Homepage: https://github.com/ged/ruby-pg
	Documentation: http://deveiate.org/code/pg
	Source Code: https://github.com/ged/ruby-pg
	Changelog: https://github.com/ged/ruby-pg/blob/master/CHANGELOG.md
	Path: /home/amc-prod/.rbenv/versions/3.3.4/lib/ruby/gems/3.3.0/gems/pg-1.6.1-x86_64-linux
```

## How to reproduce

```
$ rbenv exec bundle exec irb -rpg
/home/amc-prod/.rbenv/versions/3.3.4/lib/ruby/3.3.0/irb/init.rb:436: warning: LoadError: cannot load such file -- pg_ext
/home/amc-prod/.rbenv/versions/3.3.4/lib/ruby/3.3.0/bundled_gems.rb:74:in `require'
/home/amc-prod/.rbenv/versions/3.3.4/lib/ruby/3.3.0/bundled_gems.rb:74:in `block (2 levels) in replace_require'
/home/amc-prod/.rbenv/versions/3.3.4/lib/ruby/gems/3.3.0/gems/pg-1.6.1-x86_64-linux/lib/pg.rb:53:in `rescue in block in <module:PG>'
/home/amc-prod/.rbenv/versions/3.3.4/lib/ruby/gems/3.3.0/gems/pg-1.6.1-x86_64-linux/lib/pg.rb:47:in `block in <module:PG>'
/home/amc-prod/.rbenv/versions/3.3.4/lib/ruby/gems/3.3.0/gems/pg-1.6.1-x86_64-linux/lib/pg.rb:40:in `block in <module:PG>'
/home/amc-prod/.rbenv/versions/3.3.4/lib/ruby/gems/3.3.0/gems/pg-1.6.1-x86_64-linux/lib/pg.rb:46:in `<module:PG>'
[...]

PR #664, opened 2025-08-07
Raise a more descriptive error message for pg_ext LoadError

larskanis:
If the `3.4/pg_ext.so` doesn't load, the specific error message is overwritten by the subsequent LoadError of `pg_ext.so`. This is bad since this can hide valuable information about the load error, like a missing dependency. One such mussing dependency can be the required GLIBC version.

This patch adds a more descriptive error message containing both attempts to load the C extension. And it adds an additional help text for the special case of a GLIBC load error.

The old error:
```
cannot load such file -- pg_ext (LoadError)
```

The new error:
```
pg's C extension failed to load: (LoadError)
  /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by /root/.rubies/ruby-3.2.9/lib/ruby/gems/3.2.0/gems/pg-1.6.1-x86_64-linux/lib/3.2/pg_ext.so) - /root/.rubies/ruby-3.2.9/lib/ruby/gems/3.2.0/gems/pg-1.6.1-x86_64-linux/lib/3.2/pg_ext.so
  cannot load such file -- pg_ext

The GLIBC version of this system seems too old. Please use the source version of pg:
    gem uninstall pg --all
    gem install pg --platform ruby
or in your Gemfile:
    gem "pg", force_ruby_platform: true
See also: https://github.com/ged/ruby-pg/blob/master/README.md#source-gem
```

Changing from tabs to spaces for better message formatting.

Relates to #661

### DIFF
--- a/lib/pg.rb
+++ b/lib/pg.rb
@@ -5,141 +5,161 @@
 # The top-level PG namespace.
 module PG
 
-	# Is this file part of a fat binary gem with bundled libpq?
-	# This path must be enabled by add_dll_directory on Windows.
-	gplat = Gem::Platform.local
-	bundled_libpq_path = Dir[File.expand_path("../ports/#{gplat.cpu}-#{gplat.os}*/lib", __dir__)].first
-	if bundled_libpq_path
-		POSTGRESQL_LIB_PATH = bundled_libpq_path
-	else
-		# Try to load libpq path as found by extconf.rb
-		begin
-			require "pg/postgresql_lib_path"
-		rescue LoadError
-			# rake-compiler doesn't use regular "make install", but uses it's own install tasks.
-			# It therefore doesn't copy pg/postgresql_lib_path.rb in case of "rake compile".
-			POSTGRESQL_LIB_PATH = false
-		end
-	end
-	POSTGRESQL_LIB_PATH.freeze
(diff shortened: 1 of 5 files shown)


lib/pg.rb
```diff
@@ -5,141 +5,161 @@
...
+        EOT
+        if msg =~ /GLIBC/
+          msg += <<~EOT
+
+            The GLIBC version of this system seems too old. Please use the source version of pg:
+                gem uninstall pg --all
+                gem install pg --platform ruby
+            or in your Gemfile:
+                gem "pg", force_ruby_platform: true
+            See also: https://github.com/ged/ruby-pg/blob/master/README.md#source-gem
```
cbandy, line 70: In URLs, the `-` branch resolves to the HEAD branch, ensuring that this link will continue to work even if the HEAD branch changes its name.

```suggestion
            See also: https://github.com/ged/ruby-pg/blob/-/README.md#source-gem
```

larskanis: Sounds good! Alternatively maybe a link to https://deveiate.org/code/pg/README_md.html#label-Source+gem ? Don't have a strong opinion.

PR #668, opened 2025-08-29
Remove absolute paths from binaries which point to build directories

### LINKED ISSUE #666: pg_service.conf no longer searched for in a good default
In versions <= 1.5.9, pg_service.conf is searched for in ~/, followed by /etc/postgres-common/.

In newer versions, the fallback for ~/ is /home/lars/comcard/ruby-pg/ports/aarch64-linux/etc/postgresql/. That is less than ideal. :)

larskanis:
There are several absolute paths built into the binary. For instance fallback path to `pg_service.conf` or the fallback path to kerberos ccache file or the rpath included into the libpq or C-ext files.

They can be avoided or fixed by disabling rpaths and by using `DESTDIR` instead of `--prefix` configure option.

Fixes #666

The remaining rpath to libruby will probably be fixed by: https://github.com/rake-compiler/rake-compiler-dock/pull/165

Also fix the `MAKEFLAGS` option. `MAKEOPTS` was wrong.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -141,7 +141,7 @@ CrossLibraries.each do |xlib|
 			bundle install --local &&
 			#{ "rake install_darwin_mig[__arm64__]" if platform =~ /arm64-darwin/ }
 			#{ "rake install_darwin_mig[__x86_64__]" if platform =~ /x86_64-darwin/ }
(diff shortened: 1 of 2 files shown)


larskanis:
Related tests are added here: https://github.com/ged/ruby-pg/pull/669

PR #700, opened 2026-02-13
Add per connection support for OAuth 2.0 with PostgreSQL 18

larskanis:
This is based on #693, but changed it to a hook which is set as a connection parameter.

I plan to make it based on `st_table` instead of handling in ruby.
That should make it Ractor compatible.
Let's see what looks better...

### DIFF
--- a/ext/pg_auth_hooks.c
+++ b/ext/pg_auth_hooks.c
@@ -8,49 +8,10 @@
 
 #ifdef LIBPQ_HAS_PROMPT_OAUTH_DEVICE
 
-#ifdef TRUFFLERUBY
-static VALUE auth_data_hook;
-#else
 /*
- * On Ruby verisons which support Ractors we store the global callback once
- * per Ractor.
+ * We store the pgconn pointers in a register to retrieve the PG::Connection VALUE in the oauth hook.
  */
-#include "ruby/ractor.h"
-static rb_ractor_local_key_t auth_data_hook_key;
-#endif
-
-static void
-auth_data_hook_init(void)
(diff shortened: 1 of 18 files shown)


ext/pg_auth_hooks.c
```diff
@@ -0,0 +1,367 @@
...
+	rb_define_method(rb_cPromptOAuthDevice, "verification_uri_complete", pg_prompt_oauth_device_verification_uri_complete, 0);
+	rb_define_method(rb_cPromptOAuthDevice, "expires_in", pg_prompt_oauth_device_expires_in, 0);
+
+	rb_cOAuthBearerRequest = rb_define_class_under(rb_mPG, "OAuthBearerRequest", rb_cObject);
+	rb_undef_alloc_func(rb_cOAuthBearerRequest);
+
+	rb_define_method(rb_cOAuthBearerRequest, "openid_configuration", pg_oauth_bearer_request_openid_configuration, 0);
+	rb_define_method(rb_cOAuthBearerRequest, "scope", pg_oauth_bearer_request_scope, 0);
+	rb_define_method(rb_cOAuthBearerRequest, "token=", pg_oauth_bearer_request_token_set, 1);
+	rb_define_method(rb_cOAuthBearerRequest, "token", pg_oauth_bearer_request_token_get, 0);
```
cbandy, line 360: IMV, this hook will be used more than the device prompt. The latter requires the additional libpq library, which already has a prompt implementation.

If the device prompt is challenging in _any_ way, I recommend deferring it until needed.

PR #701, opened 2026-02-24
Add support for lambda or Proc :password key in connection Hash

jawj:
This short PR adds support for a lambda or Proc `:password` value in Postgres connection parameters.

This is useful for services like RDS or Lakebase that can be configured to require short-lived tokens as Postgres passwords.

For precedent, three major JS drivers (node-postgres, postgres.js, Bun.SQL) support the equivalent feature, passing a JS function as the password. For example: https://github.com/brianc/node-postgres/pull/1926

It's certainly possible I've gone about this the wrong way for ruby-pg but, if so, perhaps we can figure out the right way!

### DIFF
--- a/lib/pg/connection.rb
+++ b/lib/pg/connection.rb
@@ -58,11 +58,15 @@ class PG::Connection
 	# * URI string
 	# * URI object
 	# * positional arguments
+  #
+  # If an option Hash is used, :password can be a Proc or lambda that returns a
+  # string, so as to support dynamic passwords.
 	#
 	# The method adds the option "fallback_application_name" if it isn't already set.
 	# It returns a connection string with "key=value" pairs.
 	def self.parse_connect_args( *args )
 		hash_arg = args.last.is_a?( Hash ) ? args.pop.transform_keys(&:to_sym) : {}
+		hash_arg[:password] = hash_arg[:password].call if hash_arg[:password].respond_to?(:call)
 		iopts = {}
 
 		if args.length == 1
--- a/spec/pg/connection_spec.rb
+++ b/spec/pg/connection_spec.rb
@@ -313,6 +313,28 @@ describe PG::Connection do
 				PG::Connection.class_eval("PROGRAM_NAME=PG.make_shareable(#{old_script_name.inspect})")
 			end
 		end
+
+		it "accepts a lambda as the :password key of a Hash parameter" do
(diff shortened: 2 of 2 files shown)


larskanis:
IMHO the benefit of using the lambda password is zero:

```
pr = proc{ "password" }
PG.connect(password: pr)
```

compared to:

```
pr = proc{ "password" }
PG.connect(password: pr.call)
```

The only difference is at `PG::Connection#reset`, which would refresh the password from the lambda. But `#reset` is just for convenience. It can equally be replaced by a new `PG.connect`.

Is it the `#reset` function you want to support with refreshed password? Your description doesn't describe a clear reason, neither the linked issue does.

jawj:
In this simple case where you're manually creating a single connection, I guess you're right: the key benefit is that `PG::Connection#reset` works properly. That's non-zero, but certainly small.

The key scenario where a Proc or lambda for the `:password` key is useful is when using a connection pool, such as provided by ActiveRecord, Sequel, or the connection_pool gem.

A pool connects (and disconnects) clients on-demand, potentially over long periods between app restarts, using connection parameters supplied at app startup. That makes them tricky to use with the short-lived password tokens that may be provided by e.g. RDS, Google Cloud SQL, Azure or Lakebase — unless the Postgres password can be specified dynamically.

It would instead be possible to build in support for dynamic passwords separately in every pooler, I suppose. But it seems much more efficient and consistent to do it in the driver itself. And it don't think it needs to add a lot of code to maintain.

jawj:
I should maybe also clarify that the value of something like `pr = proc { "password" }` is nil, of course.

In practice, a Proc or lambda is going to be used to retrieve a token from a local cache, refresh it via an API over HTTPS, and so on.

And it's quite nice to be able encapsulate this sort of logic. For example, a Postgres provider's SDK can then supply connection parameters that include a dynamic password, and the developer using the SDK doesn't have to worry about token refresh: they can just use these connection parameters anywhere in the same way they normally would.

jawj:
Could I ask you again to consider merging this tiny PR?

frederikspang:
This would indirectly ease implementing AWS IAM for RDS, and Entra Auth for Azure Postgres, and GCP CloudSQL/IAM. (And I guess Kerberos/gssapi/etc.)

I for one would love not to do monkey patching for Azure AKS workload identity to work ... 😃

PR #704, opened 2026-05-16
Fix incomplete transaction commit when thread is shutdown ungracefully

intale:
### Problem description

Running transactions may suddenly commit when the thread it is running from is force-shutdown. 

### Steps to reproduce

Refer to the newly added test case

### Solution

To fix this we have to additionally check the thread status in `ensure` block and rollback in case we've been interrupted.

### DIFF
--- a/lib/pg/connection.rb
+++ b/lib/pg/connection.rb
@@ -306,30 +306,19 @@ class PG::Connection
 	# and a +COMMIT+ at the end of the block, or
 	# +ROLLBACK+ if any exception occurs.
 	def transaction
-		rollback = false
 		exec "BEGIN"
 		yield(self)
-	rescue PG::RollbackTransaction
-		rollback = true
-		perform_rollback
-	rescue Exception
-		rollback = true
-		perform_rollback
+	rescue PG::RollbackTransaction => error
+	rescue Exception => error
 		raise
 	ensure
-		unless rollback
-			if Thread.current.status == "aborting"
-				perform_rollback
-			else
-				exec("COMMIT")
-			end
+		if error || Thread.current.status == "aborting"
+			cancel if transaction_status == PG::PQTRANS_ACTIVE
+			block
+			exec("ROLLBACK")
+		else
+			exec("COMMIT")
 		end
-	end
-
-	private def perform_rollback
-		cancel if transaction_status == PG::PQTRANS_ACTIVE
-		block
-		exec("ROLLBACK")
 	end
 
 	### Returns an array of Hashes with connection defaults. See ::conndefaults
--- a/lib/pg/connection.rb
+++ b/lib/pg/connection.rb
@@ -311,17 +311,25 @@ class PG::Connection
 		yield(self)
 	rescue PG::RollbackTransaction
 		rollback = true
-		cancel if transaction_status == PG::PQTRANS_ACTIVE
-		block
-		exec "ROLLBACK"
+		perform_rollback
 	rescue Exception
 		rollback = true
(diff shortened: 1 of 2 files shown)


intale:
It seems I need to look into the behaviour of truffleruby as, it seems, it requires special attention

spec/pg/connection_spec.rb
```diff
@@ -1072,8 +1072,41 @@
...
+        end
+
+        context "when current thread gets halted in the middle of the running transaction" do
+
+            it "rollbacks the transaction" do
+                # abort the per-example transaction so we can test our own
+                @conn.exec('ROLLBACK')
+                @conn.exec("CREATE TABLE pie ( flavor TEXT )")
+
+                begin
```
larskanis: q = Queue.new

spec/pg/connection_spec.rb
```diff
@@ -1072,8 +1072,41 @@
...
+                    thread = Thread.new do
+                        conn = $pg_server.connect
+                        conn.transaction do
+                            conn.exec("INSERT INTO pie VALUES ('rhubarb'), ('cherry'), ('schizophrenia')")
```
larskanis: q << 1

spec/pg/connection_spec.rb
```diff
@@ -1072,8 +1072,41 @@
...
+                            # emulate some load that keeps the transaction open
+                            conn.exec("select pg_sleep(2)")
+                        end
+                    ensure
+                        conn&.close
+                    end
+                    # Wait for the thread to start
+                    sleep 1
```
larskanis: q.pop

lib/pg/connection.rb
```diff
@@ -305,24 +305,32 @@ class << self
 	# Executes a +BEGIN+ at the start of the block,
 	# and a +COMMIT+ at the end of the block, or
 	# +ROLLBACK+ if any exception occurs.
-	def transaction
```
larskanis, line 308: Could you please keep tabs indention?

larskanis (changes requested):

spec/pg/connection_spec.rb
intale: Thanks, that looks nicer.

larskanis (approved):
I added two more commits. Are they OK from your point of view?

intale:
> I added two more commits. Are they OK from your point of view?

I have a concern regarding `rescue Exception => error` case. Previously it was rescue => rollback => re-raise which was logical, but now it is rescue => re-raise => rollback. While I can't provide any edge case where this sequence change may result in any side effect - it may depend on a different ruby implementation, thus introducing some different behaviour than intended. What do you think? Do you have something in mind that could potentially affect on the final result? Otherwise looks good to me - you caught this randomly failing issue with the test I provided and replaced `pg_sleep` with `sleep`.
Imho, I would keep the implementation explicit for each case like it was before as it is more straightforward from my point of view.

larskanis:
I like the new ensure based approach more, since rollback in `rescue` makes it more cluttered. This is because `rescue` is not called in two cases: when the `yield` is aborted by `break` and when the thread is killed. But all cases are caught in the ensure branch. So it's more natural and clearer to me to handle all cases of commit or rollback there. The ruby language definition of rescue/ensure is old and well described - it's not an edge case.

intale:
> The ruby language definition of rescue/ensure is old and well described - it's not an edge case.

It is. Cool. I just tried to be paranoid about the potential behaviour. I don't have any more remarks.

larskanis:
Thank you @intale for raising this issue and providing fix and test!

PR #724, opened 2026-06-23
Update Japanese translation

gemmaro:
Hello,

This updates Japanese translation of readme, and removes the POT file from version control.

Thank you,


larskanis:
Thank you very much!

PR #726, opened 2026-06-28
Suggestion: PG::Connection#embed_params

intale:
Introduce `PG::Connection#complile` that take sql string and position parameters and compiles it into a plain sql. I couldn't find another way how to do it using built-in tools. The motivation behind this implementation:

- debugging sql queries. When you have queries with dozens(or even hundreds) of positional parameters - it is pretty cancer to replace each positional param by hands for later inspection
- such feature as [PostreSQL view](https://www.postgresql.org/docs/current/sql-createview.html) does not support positional parameters and requires pre-computed sql string. So, imaging you have some sql builder in your app that results in sql with positional params. And now you need to work with PostgreSQL view which does not support them. You will end up implementing very similar to the changes in this PR or somehow compute an sql string based on your app's logic. Both variants would be annoying.

The downside of this implementation is that it can't reliably encode binary data into a string representation except some common cases like with booleans, but do we need them at all?

### DIFF
--- a/lib/pg/connection.rb
+++ b/lib/pg/connection.rb
@@ -669,6 +669,68 @@ class PG::Connection
 	end
 	alias async_cancel cancel
 
+	PLACEHOLDER_RE = /
+		'(?:''|[^'])*'                                             | # string literal
+		"(?:""|[^"])*"                                             | # quoted identifier
+		--[^\n]*                                                   | # line comment
+		\/\*.*?\*\/                                                | # block comment
+		\$\$.*?\$\$                                                | # dollar-quoted string. E.g. $$ $1 $$
+		\$(?<__dq_tag>[A-Za-z_][A-Za-z_0-9]*)\$.*?\$\k<__dq_tag>\$ | # named dollar-quoted string. E.g. $foo$ $1 $foo$
+		(?<placeholder>\$(?:[1-9]\d*))                               # placeholder we are interested in
+	/mx
+	private_constant :PLACEHOLDER_RE
+
+	# call-seq:
+	#    conn.compile( sql, params ) -> String
+	#
+	# Compiles your prepared sql statement and the given positional arguments into plain sql.
+	# Example:
+	# 	res = conn.exec_params('SELECT $1 AS a, $2 AS b, $3 AS c', [1, 2, nil])
+	# 	# => "SELECT '1' AS a, '2' AS b, NULL AS c"
+	def compile(sql, params)
+		return sql if params.empty?
+
+		sql.gsub(PLACEHOLDER_RE).each do |matched|
+			placeholder = Regexp.last_match[:placeholder]
+			# Do not replace non-positional args string and pass it as is
+			next matched unless placeholder
+
+			value = params[placeholder[1..].to_i - 1]
+			value = encode_value(value)
+			normalize_value(value)
+		end
+	end
+
+	private def encode_value(value)
+		return value if type_map_for_queries.is_a?(PG::TypeMapAllStrings)
+		unless type_map_for_queries.is_a?(PG::TypeMapByClass)
+			raise <<~TEXT.strip
+				Unsupported type map. Please use the one which is inherited from PG::TypeMapByClass, for example \
+				PG::BasicTypeMapForQueries:
+				conn = PG::Connection.new
+				conn.type_map_for_queries = PG::BasicTypeMapForQueries.new(conn)
+			TEXT
+		end
+
+		encoder = type_map_for_queries[value.class]
+		return type_map_for_queries.send(encoder, value).encode(value) if encoder.is_a?(Symbol)
+		# format == 1 stands for binary format
+		return value if encoder.nil? || encoder.format == 1
+
+		encoder.encode(value)
+	end
+
+	private def normalize_value(value)
+		case value
+		when TrueClass, FalseClass
+			"#{value}"
+		when NilClass
+			'NULL'
+		else
+			"'#{self.class.escape(value.to_s)}'"
+		end
+	end
+
 	module Pollable
 		# Track the progress of the connection, waiting for the socket to become readable/writable before polling it.
 		#
--- a/lib/pg/connection.rb
+++ b/lib/pg/connection.rb
@@ -680,32 +680,66 @@ class PG::Connection
 	/mx
 	private_constant :PLACEHOLDER_RE
 
-	# call-seq:
-	#    conn.compile( sql, params ) -> String
+	# Compiles your prepared SQL statement and the given positional arguments into plain SQL string.
 	#
-	# Compiles your prepared sql statement and the given positional arguments into plain sql.
+	# The resulting SQL string can be used with +conn.exec+ like the prepared SQL statement and parameters with +conn.exec_params+.
+	# +conn.exec_params+ is usually preferred because it's faster and safer.
+	# +embed_params+ is intended for debugging messages with positional parameters.
+	# It avoids manual insertion for later inspection in +psql+ or so.
+	#
 	# Example:
-	# 	res = conn.exec_params('SELECT $1 AS a, $2 AS b, $3 AS c', [1, 2, nil])
+	# 	res = conn.embed_params('SELECT $1 AS a, $2 AS b, $3 AS c', [1, 2, nil])
 	# 	# => "SELECT '1' AS a, '2' AS b, NULL AS c"
-	def compile(sql, params)
+	def embed_params(sql, params, type_map: type_map_for_queries, coder_maps_bundle: nil)
 		return sql if params.empty?
 
-		encoders = type_map_for_queries.query_param_encoders(params)
+		oid_to_typecast = proc do |oid|
+			if oid && oid > 0
+				by_oid = if coder_maps_bundle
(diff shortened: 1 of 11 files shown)


larskanis:
That sounds like a great idea! It's because I manually inserted too many parameters into too many queries from rails debug log in order to execute it in `psql`. That way a more useful debug output in rails would be possible and that would be my intention.

larskanis:
I added a commit to use generic PG::TypeMap methods only and regarding binary data and binary format.

The method name "compile" is not very specific. Is it used in any other database driver? How sounds `embed_params` instead?

intale:
@larskanis thank you. ~Could you please explain how your changes apply to `PG::TypeMapAllStrings` as well? Because it has its own interface, all static, not exposed to public and `PG::TypeMapAllStrings`  the default class for type mapping.~ Nvm, I'm blind. `pg_tmas_type` is inherited from `pg_typemap_type` and you added changes exactly into `pg_typemap_type`. I mixed it with `PG::TypeMapByClass`. Lgtm

intale:
> The method name "compile" is not very specific. Is it used in any other database driver? How sounds embed_params instead?

I don't have any strong preferences regarding this. `embed_params` works as well

larskanis:
🎉 With the latest change the whole activerecord test suite runs through with only 2 expected errors, when I change the following AR lines to got through `embed_params` and sending the resulting SQL text without binds:
https://github.com/rails/rails/blob/09bf0878ec2a604f4f1430c8757db23d16925f14/activerecord/lib/active_record/connection_adapters/postgresql/database_statements.rb#L162-L184

intale:
@larskanis thank you for your effort and covering cases I was not aware of.

lib/pg/connection.rb
```diff
@@ -669,6 +669,89 @@ def cancel
...
+		\$\$.*?\$\$                                                | # dollar-quoted string. E.g. $$ $1 $$
+		\$(?<__dq_tag>[A-Za-z_][A-Za-z_0-9]*)\$.*?\$\k<__dq_tag>\$ | # named dollar-quoted string. E.g. $foo$ $1 $foo$
+		(?<placeholder>\$(?:[1-9]\d*))                               # placeholder we are interested in
+	/mx
+	private_constant :PLACEHOLDER_RE
+
+	# Compiles your prepared SQL statement and the given positional arguments into plain SQL string.
+	#
+	# The resulting SQL string can be used with +conn.exec+ like the prepared SQL statement and parameters with +conn.exec_params+.
+	# +conn.exec_params+ is usually preferred because it's faster and safer.
```
cbandy, line 686: IME, these kinds of not-quite-parsers are vulnerable to SQL injection.

Perhaps a stronger or more specific statement about how this should/not be used with user input.

intale: I agree. If there is any mistake in regexp - the result may end up containing sql injection. For example

```ruby
embed_params("select '$1' as one, $1", [%{', 2 as two, }])
```

may result in

```ruby
[{"?column?" => "', 2 as two, ", "two" => "2", "one" => ""}]
```

instead this

```ruby
[{"one" => "$1", "?column?" => "', 2 as two, "}]
```

if we forget to cover regular string literal(e.g. `'my str'`)  in `PLACEHOLDER_RE`. Thus, as long as we cover all possible variations of string literals and escape the value - we are safe. Fortunately I covered all possible scenarios in tests. As an additional measure - I can put here a guard that fails on new postgresql versions and a dev should check if any new string literals appear in a new postgresql version and adjust the implementation accordingly. Something like this(in `#embed_params_and_check`):

```ruby
# Query returns a string like "18.4 (Ubuntu 18.4-0ubuntu0.26.04.1)"
current_pg_version = conn.exec('SHOW server_version').first['server_version'].split(' ').first
current_pg_version = Gem::Version.new(current_pg_version)
expect(current_pg_version).to be < Gem::Version.new('19'), "PostgreSQL 19 is not tested. Please ensure there are no new forms of a string literal and increase the version to make this test pass."
```

What do you think, guys?

larskanis: Yes, maybe we should make a stronger statement about security. Putting SQL and user data together into one string is always more risky than keeping them separate.

On the other hand to exploit a incomplete or broken regex it's necessary to have a strange SQL placeholder string. But this string must not contain any unchecked user inputs anyway, so cannot be manipulated. And the `params` items go through the `escape` or `escape_bytea` methods provided by libpq, so that they should be safe.

So I think it's not necessary to put a guard for new PostgreSQL versions. But maybe someone speaking native English can improve my wording or make it more specific.

intale: Unfortunately `#escape` does not fully protect us from sql injections. This is because we have to support every possible string literals here, but `#escape` only ensures there is correct amount of `'` in the string. Thus, if some regexp contains a mistake - an injection becomes possible for other string literals. In the example bellow you can see an injection using `$$` in case [this](https://github.com/intale/ruby-pg/blob/807c5911608beeaee607e54aaead5251bf2eaa77/lib/pg/connection.rb#L677) variant is not handled properly(you can simply delete that variant to reproduce the issue):

```ruby
sql = "select $$ $1 as t, $$ as one, $1"
params = ["$$ as tt, $$"]
@conn.exec(@conn.embed_params(sql, params)).to_a
# => [{"tt" => " '", "one" => "' as t, ", "?column?" => "$$ as tt, $$"}]
@conn.exec_params(sql, params).to_a
# [{"one" => " $1 as t, ", "?column?" => "$$ as tt, $$"}]
```

Thus, such issue might appear because we basically replace all appearances of positional variable in the string thanks to `gsub` and incorrect/absent regexp. This is what I meant in my previous message when I was suggesting to add another guard against new potential string literals in future PostgreSQL versions.

larskanis: @intale I understand your concern. But what I mean is that the SQL string must be manipulated, in order to exploit a flaw in the regex, which should not be possible for an attacker. An attacker can change the `params`, but he cannot exploit a flaw in the regex.

intale: Thanks for explanation. That is true. Let me rephrase my concern. My intention of adding a guard is to prevent next case:
- current implementation covers all cases for current PostreSQL
- PosteeSQL v25 introduces new string literal. Let's say `%% my str %%`
- the implementation suddenly becomes vulnerable to sql injections if new string literal is used because it misses a regexp to handle it

But if we would have a guard that demands manual check of new PostreSQL version support before succeeding the build - we would be more safe than without it.

intale: @larskanis hey. So what do you think? Do my convulsions makes sense to you?

larskanis: Is there a documentation page the version guard could point to, to verify that expectations still apply?

larskanis:
From my point of view this implementation is complete now.

To keep things simple, I chose to revert the automatic OID => type name conversion and require explicit type names for casts. The accessor for `coder_maps_bundle` is still there, although not used. But it seems to be valuable anyway and it has simple tests already.

Is the description OK? Anything else?

lib/pg/connection.rb
intale: @larskanis I prepared PostgreSQL version that supports hypothetical `%%` string literal https://github.com/intale/postgres/tree/double_percent_literal. I also pushed a commit that demonstrates the problem https://github.com/intale/ruby-pg/commit/f44b9d5618d0ce8bfe50ac12868ed7dd0c8b599d. You should build that version of PostgreSQL and run tests like so:

```bash
PATH=/path/to/bin:$PATH rspec -Ilib -fd spec/pg/connection_spec.rb:3055
```

So, in order that test to pass we need to add support of `%%` string literal. In that example I already added it, but commented it out to demonstrate the problem.

All I am trying to say is that current implementation is permissive by default. We handle all possible string literals in `PLACEHOLDER_RE`, but any new hypothetical string literal which is not handled by that regexp would make that implementation immediately unsafe.

cbandy: @larskanis, if I understand your question correctly, this page details the different kinds of quoting in Postgres: https://www.postgresql.org/docs/current/sql-syntax-lexical.html

intale: Lars, after reading Cris's comment I realized I may misunderstood your question. Do you want to use it as some kind of reliable point to perform some assertions to make a decision regarding this implementation? As an idea - the assertion can calculate the checksum of the page Cris posted and compare it with a known checksum, thus failing if docs get adjusted.

larskanis: @cbandy Thanks, I wasn't aware of this page!
@intale I don't think we need hashing - just a server version check in the specs and a link to that syntax page would be nice IMHO.

larskanis: @intale Given that there are only a few changes in the SGML file per year, maybe your idea is better than checking the server version. So just checking the latest commit hash? 
https://github.com/postgres/postgres/commits/master/doc/src/sgml/syntax.sgml

intale: @larskanis ah, cool. Haven't been thinking we have it in git. In this case the latest commit hash of that file is just enough.

intale:
@larskanis I pushed the guard we were discussing. Please check

